PR #2, opened 2018-05-14
Bring list to parity with Redump known types

mnadareski:
This change adds all remaining consoles and disc-based devices that are currently listed in Redump as valid devices. This also does some minor renames and splits that bring this to parity as well. One minor TODO note was added possibly for future ease-of-use for dumping IBM PC software.

The second commit disables application signing because no certificate is in the repository and nobody but the original author would be able to build this from source otherwise.

### DIFF
--- a/DICUI.csproj
+++ b/DICUI.csproj
@@ -62,8 +62,9 @@
     <GenerateManifests>true</GenerateManifests>
   </PropertyGroup>
   <PropertyGroup>
-    <SignManifests>true</SignManifests>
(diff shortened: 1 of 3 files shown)


mnadareski:
One additional suggestion: with the TODO that I added, I realize that detecting the protection is not something that can be done easily through a script, so I would recommend having a single set of flags that would cover the IBM PC cases no matter what. Because dumping a disc with no SecuROM or something with the same flags as one that does will not affect how that first disc is dumped.

PR #56, opened 2018-06-18
Separated of System and Disc Type

### LINKED ISSUE #51: Separation of System and Type
Currently, the system dropdown includes both the system and the type in one fell swoop. It may be in our best interest to make these completely separate things in order to make the disc cleaner. This was recommended and coded by Jack, whose PR will be coming.

**Notes:**
- This may work best by using the already-created "get disc types for known system" method
- Having it repopulate the Disc Type dropdown each time is also probably the best idea
- Placement on the screen may be tricky, but it is doable

Jakz:
As proposed and discussed before this PR separates choosing a system from choosing a disk type in the UI so that it is managed by two different combo boxes.

Fixes #51


MainWindow.xaml.cs
```diff
@@ -133,16 +140,47 @@ private void txt_OutputDirectory_TextChanged(object sender, TextChangedEventArgs
...
+        private void PopulateDiscTypeAccordingToChosenSystem()
+        {
+          cmb_DiscType.SelectedIndex = -1;
+
+          var currentSystem = cmb_SystemType.SelectedItem as Tuple<string, KnownSystem?>;
+
+          if (currentSystem != null && currentSystem.Item2 != KnownSystem.NONE)
+          {
+
+            List<Tuple<string, DiscType?>> allowedDiscTypesForSystem = Utilities.Validation.GetValidDiscTypes(currentSystem.Item2)
```
mnadareski: Why not just modify what comes out of GetValidDiscTypes? It seems that would make the code cleaner.

MainWindow.xaml.cs
```diff
@@ -24,7 +24,7 @@ public partial class MainWindow : Window
         // Private UI-related variables
         private List<Tuple<char, string, bool>> _drives { get; set; }
         private List<int> _driveSpeeds { get { return new List<int> { 1, 2, 3, 4, 6, 8, 12, 16, 20, 24, 32, 40, 44, 48, 52, 56, 72 }; } }
-        private List<Tuple<string, KnownSystem?, DiscType?>> _systems { get; set; }
+        private List<Tuple<string, KnownSystem?>> _systems { get; set; }
```
mnadareski: For the time being, can you also have a persistent _discTypes list here as well? That helps keep track of what is being used globally.

MainWindow.xaml.cs
```diff
@@ -230,7 +268,8 @@ private void BrowseFolder()
             // If we have a custom configuration, we need to extract the best possible information from it
             if (systemName == "Custom Input" && system == KnownSystem.NONE && type == DiscType.NONE)
             {
-                Utilities.Validation.DetermineFlags(customParameters, out type, out system, out string letter, out string path);
+                string letter, path;
+                Utilities.Validation.DetermineFlags(customParameters, out type, out system, out letter, out path);
```
mnadareski: I know you had problems with the older version of VS and this, but can you put this back to inline variable declaration? It's much cleaner that way.

MainWindow.xaml.cs
```diff
@@ -267,7 +306,7 @@ private void BrowseFolder()
                     {
                         FileName = dicPath,
                         Arguments = parameters,
-                    },
+                    }
```
mnadareski: This is a pedantic nit: I leave the dangling commas so that future changes touch less code. It's something I've had to do for work so it's built in. If you don't mind returning that, I'd appreciate it.

MainWindow.xaml.cs
```diff
@@ -446,64 +485,77 @@ private void CancelDumping()
...
                         btn_StartStop.IsEnabled = false;
-                    }
-                    else
-                    {
-                        lbl_Status.Content = string.Format("{0} ready to dump", Converters.DiscTypeToString(tuple.Item3));
+                        break;
+                    case DiscType.GameCubeGameDisc:
+                    case DiscType.GDROM:
+                        lbl_Status.Content = string.Format("{0} discs are partially supported by DIC", Converters.DiscTypeToString(selectedDiscType));
                         btn_StartStop.IsEnabled = (_drives.Count > 0 ? true : false);
```
mnadareski: Wouldn't the DiscTypeToString calls return the same as discTypeTuple.Item1? That would reduce the number of method calls here.

MainWindow.xaml.cs
```diff
@@ -446,64 +485,77 @@ private void CancelDumping()
...
+
+            //TODO: a specific KnownSystem to manage Custom would be better
+            bool isCustom = systemTuple.Item1 == "Custom Input";
+            var selectedSystem = systemTuple.Item2;
+            var selectedDiscType = discTypeTuple != null ? discTypeTuple.Item2 : DiscType.NONE;
+
+            // No system chosen, update status
+            if (selectedSystem == KnownSystem.NONE)
             {
-                cmb_DiscType.SelectedIndex++;
```
mnadareski: Why did you separate out the KnownSystem.NONE from the rest of the switch statement? I don't see any advantage to doing so

MainWindow.xaml.cs
```diff
@@ -515,7 +567,7 @@ private void EnsureDiscInformation()
             }
 
             // Special case for Custom input
-            if (tuple.Item1 == "Custom Input" && tuple.Item2 == KnownSystem.NONE && tuple.Item3 == DiscType.NONE)
+            if (isCustom && selectedSystem == KnownSystem.NONE && selectedDiscType == DiscType.NONE)
```
mnadareski: If you're making `isCustom` its own variable, why not wrap in the other two parts of this conditional into that one boolean?

MainWindow.xaml.cs
```diff
@@ -619,7 +671,8 @@ private void SetSupportedDriveSpeed()
 
             int index = output.IndexOf("ReadSpeedMaximum:");
             string readspeed = Regex.Match(output.Substring(index), @"ReadSpeedMaximum: [0-9]+KB/sec \(([0-9]*)x\)").Groups[1].Value;
-            if (!Int32.TryParse(readspeed, out int speed))
+            int speed;
+            if (!Int32.TryParse(readspeed, out speed))
```
mnadareski: Ditto here with the inlining

MainWindow.xaml.cs
```diff
@@ -569,7 +621,7 @@ private void EnsureDiscInformation()
         private void GetOutputNames()
         {
             var driveTuple = cmb_DriveLetter.SelectedItem as Tuple<char, string, bool>;
-            var discTuple = cmb_DiscType.SelectedItem as Tuple<string, KnownSystem?, DiscType?>;
+            var discTuple = cmb_SystemType.SelectedItem as Tuple<string, KnownSystem?, DiscType?>;
```
mnadareski: This one needs to be `Tuple<string, KnownSystem?>`

Utilities/DumpInformation.cs
```diff
@@ -186,7 +186,8 @@ public static bool FoundAllFiles(string outputDirectory, string outputFilename,
                             break;
                         case KnownSystem.SegaSaturn:
                             mappings[Template.SaturnHeaderField] = GetSaturnHeader(GetFirstTrack(outputDirectory, outputFilename)).ToString();
-                            if (GetSaturnBuildInfo(mappings[Template.SaturnHeaderField], out string serial, out string version, out string buildDate))
+                            string serial, version, buildDate;
+                            if (GetSaturnBuildInfo(mappings[Template.SaturnHeaderField], out serial, out version, out buildDate))
```
mnadareski: Ditto here with the inlining

Utilities/DumpInformation.cs
```diff
@@ -271,7 +272,8 @@ public static bool FoundAllFiles(string outputDirectory, string outputFilename,
                         case KnownSystem.MicrosoftXBOX:
                         case KnownSystem.MicrosoftXBOX360XDG2:
                         case KnownSystem.MicrosoftXBOX360XDG3:
-                            if (GetXBOXAuxInfo(combinedBase + "_disc.txt", out string dmihash, out string pfihash, out string sshash, out string ss))
+                            string dmihash, pfihash, sshash, ss;
+                            if (GetXBOXAuxInfo(combinedBase + "_disc.txt", out dmihash, out pfihash, out sshash, out ss))
```
mnadareski: Ditto here with the inlining

Utilities/Validation.cs
```diff
@@ -1033,7 +1017,8 @@ private static bool DoesNextExist(List<string> parameters, int index)
         /// <returns>True if it's a valid number, false otherwise</returns>
         private static bool IsValidNumber(string parameter, int lowerBound = -1, int upperBound = -1)
         {
-            if (!Int32.TryParse(parameter, out int temp))
+            int temp;
+            if (!Int32.TryParse(parameter, out temp))
```
mnadareski: Ditto here with the inlining

MainWindow.xaml.cs
Jakz: The switch is over DiscType, not on KnownSystem, we're comparing apples with oranges here.

MainWindow.xaml.cs
Jakz: Sure I will but I don't guarantee to remember to put them where needed since it's not in my code taste.

MainWindow.xaml.cs
Jakz: Sure, I'll build a List<DiscType?> of all currently used types for the systems, which sounds like what we're looking for since it's just to track it.

MainWindow.xaml.cs
Jakz: Yes, that's a side effect of the refactor of the top part for the new management, just haven't modified it, but this `discTypeTuple.Item1` is so unreadable, damn unnamed tuples.

MainWindow.xaml.cs
Jakz: Actually we don't need to fetch anything again here since we already have `selectedSystem` and `selectedDiscType` from above, I'll enhance accordingly.

MainWindow.xaml.cs
Jakz: Actually I don't like `isCustom` neither, I'll just define a new `KnownSystem.CUSTOM` for this.

Jakz:
Ok, I should have pushed all relevant issues and tweaks.

MainWindow.xaml.cs
mnadareski: This would still be incorrect as `cmb_SystemType.SelectedItem` is a `Tuple<string, KnownSystem?>` and this would return null always.

Utilities/Validation.cs
```diff
@@ -419,49 +418,46 @@ public static class Validation
...
+        {
+            return systems
+                .ConvertAll(s => GetValidDiscTypes(s))
+                .SelectMany(d => d)
+                .Distinct()
+                .Select(d => Tuple.Create(Converters.DiscTypeToString(d), d))
+                .OrderBy(t => t.Item1)
+                .ToList();
+
+        }
```
mnadareski: Nit: Remove any extra space in the method. Not functional, just a style thing.

MainWindow.xaml.cs
Jakz: yes, I was actually talking about something similar in `EnsureDiscInformation()` method, here we need to fetch them.

PR #58, opened 2018-06-19
Formats and Types

### LINKED ISSUE #52: DVD5/9, BD25/50 combination
At some point, it may be worthwhile to figure out how to combine the single- and dual-layer entries for the disc formats that support them so there's less confusion for end users. Maybe have the output say which it ended up being.

**Ideas:**
- Use the `GetLayerbreak` method to see if any layerbreak is returned. If one is returned, then it's dual layer, otherwise, single layer. This could be THE key.
- Keep both internal options but only expose one of them at any given time to end users (not sure if I like this idea, but it's possible). It would also be one of those "if both exist, only show one"... which could run into potential issues

### LINKED ISSUE #53: Format additions/edits
HD-DVD should be removed from the XBOX360 format list because no games ever came out on it. However, since movies did, it's technically good to keep them their own format, similar to DVD-Video. Additionally, adding currently undumpable formats such as Laserdisc to the list may be fun to do, as long as the implementer makes sure to have the status of those formats be undumpable. See current implementation for HD-DVD and UMD.

mnadareski:
This PR addresses the following:

- Fixes #52 - This needs testing, but I have combined the single- and dual-layer disc types into a single enum value each. This is achieved by checking for the presence of the layerbreak field in the output data. At the moment, this means that it is entirely transparent to the user what type of disc this is. Followup item (possibly for this PR): Output the disc type to either the name of the file OR a field within the submission info file.
- Fixes #53 - A few more undumpable formats have been added to the internal list along with the migration of HD-DVD out of the valid formats for X360 (XDG2/3). This should address any semantic issues as well as allow for some fairly interesting cases later when more options arise besides DIC to backup media.
- Cleanup and bugfixes for the split system and disc types. The code was good, I just had a couple of cleanups to reduce the amount of duplicate code and one null fix that was missed in the initial PR.

### DIFF
--- a/Constants.cs
+++ b/Constants.cs
@@ -95,8 +95,8 @@
         public const string DATField = "DAT";
         public const string ErrorCountField = "Error Count";
         public const string CuesheetField = "Cuesheet";
-        public const string SubIntentionField = "SubIntention (SecuROM)";
-        public const string WriteOffsetField = "WriteOffset";
+        public const string SubIntentionField = "SubIntention Data (SecuROM)";
+        public const string WriteOffsetField = "Write Offset";
         public const string LayerbreakField = "Layerbreak";
         public const string PlaystationEXEDateField = "EXE Date";
         public const string PlayStationEDCField = "EDC"; // TODO: Not automatic yet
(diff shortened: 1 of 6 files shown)


mnadareski:
Finally confirmed that the layerbreak check acts as intended by ripping a DVD-9 and observing the results. This is now being checked in.

PR #63, opened 2018-06-20
Enhancement of options management

Jakz:
This pull request refactors the management of current path properties:

- properties are moved in a dedicated `Options` file which also manage persistence
- the handmade UI for options is removed and a dedicated `OptionsWindow.xaml` file is created
- all the current interactions are bridged through new options management

This refactor is not meant to be final since `Options` currently relies on reflection for a fast replacement solution of actual management, but we'll fix this successively.


MainWindow.xaml
```diff
@@ -21,7 +21,7 @@
 
         <ToolBarTray Grid.Row="0" Grid.Column="0" Grid.ColumnSpan="2" Height="30" HorizontalAlignment="Stretch" IsLocked="True">
             <ToolBar>
-                <Button x:Name="tbr_Properties" Content="Properties" Command="Properties" Click="tbr_Properties_Click" CommandManager.CanExecute="tbr_Properties_CanExecute"/>
+                <Button x:Name="tbr_Options" Content="Options" Click="tbr_Options_Click" CommandManager.CanExecute="tbr_Options_CanExecute"/>
```
mnadareski: I think that the `Command="Properties"` was required in some way, but I cannot confirm why.

MainWindow.xaml.cs
```diff
@@ -105,26 +101,23 @@ private void cmb_DriveSpeed_SelectionChanged(object sender, SelectionChangedEven
...
+            // lazy initialization
+            if (_optionsWindow == null)
+                _optionsWindow = new OptionsWindow(_options);
 
-        private void tbr_Properties_CanExecute(object sender, System.Windows.Input.CanExecuteRoutedEventArgs e)
-        {
-            e.CanExecute = true;
-        }
+            //_optionsWindow.Owner = this;
+            //_optionsWindow.WindowStartupLocation = WindowStartupLocation.CenterOwner;
```
mnadareski: Why are these two lines commented out? They seem like pretty valid options to enable.

MainWindow.xaml.cs
```diff
@@ -240,6 +233,11 @@ private void BrowseFolder()
         /// </summary>
         private async void StartDumping()
         {
+            string dicPath = _options.dicPath;
+            string sgRawPath = _options.sgRawPath;
+            string psxtPath = _options.psxtPath;
+            string subdumpPath = _options.subdumpPath;
```
mnadareski: No need to set these here if you just call the `_options.blahPath` variables below.

MainWindow.xaml.cs
```diff
@@ -644,8 +642,9 @@ private void SetSupportedDriveSpeed()
                 return;
             }
 
-            // Validate that the required program exits
-            if (!File.Exists(dicPath))
+            // Validate that the required program exits and it's not DICUI itself
```
mnadareski: Nit: I can't spell apparently... If you can change `program exits` to `program exists`, I'd appreciate it

mnadareski:
I'll have to test to see how the new window plays out with saving and loading, but the couple small comments are all I could see.

MainWindow.xaml.cs
Jakz: Actually I did this to avoid renaming them all and keep the following code more readable. I'd avoid making string copies if it's possible in C# (eg. `ref string dicPath = ref _options.dicPath`) but it's a micro optimization.

MainWindow.xaml.cs
Jakz: I commented them because they didn't work as expected but probably the problem was somewhere else.

MainWindow.xaml.cs
mnadareski: In the rest of the code you still reference the `_options.` ones, though. So if you want to make the least changes, I'd keep those variables more global and just have the setter there. Then you shouldn't have to worry. Or go the other way and remove the usage of those variables completely.

Jakz: Yes, that's because we're dealing with `StartDumping()` function which is the big boy and each path is potentially (and actually) required multiple times, this is not true in the single cases around for which aliasing the name was just redundant.

mnadareski: Okay, that makes sense in this case. Looks good, just need to test it.

mnadareski:
I was able to test locally. The new options box looks much cleaner than it did before. I also confirmed that the saving and loading of options was consistent. This is going to be a nit, but later on, it might be cool to have the output directory automatically update if you change the default. This isn't necessary for this change, just something fun for the future.

PR #76, opened 2018-06-26
Refactor of dump procedure into smaller components

Jakz:
This PR includes some refactors aimed to have a more modular code and be able to integrate the stdout of DIC process inside DICUI itself in the short future.

- `Tasks` class has been created to implement all the dump single tasks
- `Result` class has been created to managed a generic success/failure value with optional message and convertble to `bool`
- extensions have been added for `MediaType` and `KnownSystem` enums to reduce the amount of static function calls around (and keep code more object oriented)
- `StartDumping` has been split into many smaller chunks of code which are then invoked one after the other
- `EnsureDiscInformation` has been split into `EnsureCorrectInformationForSystemAndMediaType` which takes care of checking system and media type
- working proof of concept: `cmb_MediaType` combobox model now stores a `List<MediaType?>` instead that a `List<KeyValuePair<string, MediaType>>`, the display name is now rendered through a converter which is bound to the data template of the combobox itself. This simplifies the code a lot and it's extensible to KnownSystem too in the short term if we like it


MainWindow.xaml.cs
```diff
@@ -21,12 +21,15 @@ public partial class MainWindow : Window
...
-        private Process childProcess { get; set; }
+        private List<MediaType?> _mediaTypes { get; set; }
+        //private Process childProcess { get; set; }
 
         private OptionsWindow _optionsWindow;
         private Options _options;
 
+        private DumpEnvironment _env;
+
+
```
mnadareski: Nit: Since we have a lot of locals now, might be worth reordering them and grouping them for clarification.
Secondary even more minor nit: generally don't have double lines between stuff. If grouping is needed, we can use the `#region` and `#endregion` tags

MainWindow.xaml.cs
```diff
@@ -21,12 +21,15 @@ public partial class MainWindow : Window
         private MediaType? _currentMediaType { get; set; }
         private List<int> _driveSpeeds { get { return new List<int> { 1, 2, 3, 4, 6, 8, 12, 16, 20, 24, 32, 40, 44, 48, 52, 56, 72 }; } }
         private List<KeyValuePair<string, KnownSystem?>> _systems { get; set; }
-        private List<KeyValuePair<string, MediaType?>> _mediaTypes { get; set; }
-        private Process childProcess { get; set; }
+        private List<MediaType?> _mediaTypes { get; set; }
+        //private Process childProcess { get; set; }
```
mnadareski: Any particular reason why childProcess is being kept around? Or will that be removed in a future update?

MainWindow.xaml.cs
```diff
@@ -233,236 +233,52 @@ private void BrowseFolder()
...
-        /// </summary>
-        private async void StartDumping()
+        private DumpEnvironment DetermineEnvironment()
         {
-            btn_StartStop.Content = UIElements.StopDumping;
+            DumpEnvironment env = new DumpEnvironment();
+
+            // Paths to tools
+            env.subdumpPath = _options.subdumpPath;
+            env.psxtPath = _options.psxtPath;
```
mnadareski: After my change goes in, all references to `psxt` can go away 👍

MainWindow.xaml.cs
```diff
@@ -612,19 +350,79 @@ private void EnsureDiscInformation()
             }
         }
 
+        /// <summary>
+        /// 
+        /// </summary>
+        private Result EnsureCorrectInformationForSystemAndMediaType(KnownSystem? system, MediaType? type)
```
mnadareski: Maybe rename to something like `GetSupportStatus`? The full info can go in the summary

Tasks.cs
```diff
@@ -0,0 +1,302 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows;
+
+namespace DICUI
+{
+    class Result
```
mnadareski: IIRC, default visibility of classes is `internal`. It's not an official thing, but I tend to like to explicitly put access modifiers on everything. In this case, I'd call it `public`. Not necessary, but fun.

Tasks.cs
```diff
@@ -0,0 +1,302 @@
...
+        public static Result Success(string message) => new Result(true, message);
+        public static Result Success(string message, params object[] args) => new Result(true, string.Format(message, args));
+
+        public static Result Failure(string message) => new Result(false, message);
+        public static Result Failure(string message, params object[] args) => new Result(false, string.Format(message, args));
+
+        public static implicit operator bool(Result result) => result.success;
+    }
+
+    class DumpEnvironment
```
mnadareski: Ditto on access modifier here.

Utilities/Converters.cs
```diff
@@ -1,9 +1,61 @@
 ﻿using System.Collections.Generic;
 using IMAPI2;
 using DICUI.Data;
+using System.Windows.Data;
+using System;
+using System.ComponentModel;
+using System.Reflection;
+using System.Globalization;
```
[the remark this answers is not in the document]
Jakz: I never add an using directive by hand so I guess MSVS did it wrong.

Utilities/Converters.cs
```diff
@@ -1,9 +1,61 @@
...
+
+        public static bool DoesSupportDriveSpeed(this MediaType? type)
+        {
+            return type != MediaType.BluRay && type != MediaType.Floppy;
+        }
+    }
+
+    public static class KnownSystemExtensions
+    {
+        public static bool DoesSupportDriveSpeed(this KnownSystem? system)
```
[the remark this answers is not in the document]
Jakz: yes the idea was to move Name (and all related infos) but we don't need to do it at once. I was waiting to see if we all like the change.

mnadareski:
Looks good!

PR #77, opened 2018-06-28
Added preferred dump speed option

Jakz:
Added preferred dump speed option for CD-ROM and DVD-ROM to OptionsWindow. The change had not been fully integrated in current workflow as we're still evolving how this should interact with maximum drive speed and media type.

All data structures, options persistence and retrieval are ready to be used though.


Data/Constants.cs
```diff
@@ -1,13 +1,45 @@
...
+using System.Linq;
+using System.Windows.Media;
+
+namespace DICUI.Data
 {
     /// <summary>
-    /// Text for UI elements
+    /// Variables for UI elements
     /// </summary>
     public static class UIElements
```
mnadareski: Nitpick: If there is more than just constant values in here now, I'd suggest moving UIElements to either its own file or to another, more matching one

MainWindow.xaml.cs
```diff
@@ -504,7 +512,7 @@ private void SetCurrentDiscType()
             }
 
             // Get the current optical disc type
-            _currentMediaType = Validators.GetDiscType(selected?.Key);
+            _currentMediaType = null;// Validators.GetDiscType(selected?.Key);
```
mnadareski: Why is this being commented out now? I don't see _currentMediaType being set here anymore... which makes the entire next "If we have an invalid current type" section always call

Options.cs
```diff
@@ -46,5 +60,15 @@ public string Get(string key)
...
             return GetType().GetProperty(key, BindingFlags.Public | BindingFlags.Instance).GetValue(this) as string;
         }
+
+        public int GetPreferredDumpSpeedForMediaType(MediaType? type)
+        {
+            switch (type)
+            {
+                case MediaType.CD: return preferredDumpSpeedCD;
+                case MediaType.DVD: return preferredDumpSpeedDVD;
+                default: return 1; //TODO: think what we want here
```
mnadareski: The "reasonable default" we use right now is 8, I think that would be decent for this as well.

OptionsWindow.xaml.cs
```diff
@@ -94,22 +96,35 @@ private void btn_BrowseForPath_Click(object sender, EventArgs e)
...
+        {
+            Array.ForEach(PathSettings(), setting => TextBoxForPathSetting(setting).Text = _options.Get(setting));
+
+            slider_DumpSpeedCD.Value = _options.preferredDumpSpeedCD;
+            slider_DumpSpeedDVD.Value = _options.preferredDumpSpeedDVD;
+        }
+
+        #region Event Handlers
+
+        private void OnAcceptClick(object sender, EventArgs e)
```
mnadareski: Any particular reason you renamed this from `btn_Accept_Click`? It's not a rule, but in general, UI callbacks in C# are named after the UI elements that call them, since there can be multiple.

OptionsWindow.xaml.cs
```diff
@@ -94,22 +96,35 @@ private void btn_BrowseForPath_Click(object sender, EventArgs e)
...
+            _options.preferredDumpSpeedDVD = Convert.ToInt32(slider_DumpSpeedDVD.Value);
+
             _options.Save();
             Hide();
+
+            _mainWindow.OnOptionsUpdated();
         }
 
-        private void btn_Cancel_Click(object sender, EventArgs e)
+        private void OnCancelClick(object sender, EventArgs e)
```
mnadareski: Same here as `btn_Accept_Click`

MainWindow.xaml.cs
```diff
@@ -113,7 +112,7 @@ private void tbr_Options_Click(object sender, RoutedEventArgs e)
             // lazy initialization
             if (_optionsWindow == null)
             {
-                _optionsWindow = new OptionsWindow(_options);
+                _optionsWindow = new OptionsWindow(this, _options);
```
mnadareski: Clarification: Why is `this` needed here now? Is that the parenting aspect?

MainWindow.xaml.cs
```diff
@@ -210,8 +215,9 @@ private void PopulateDrives()
         /// </summary>
         private void PopulateDriveSpeeds()
         {
-            cmb_DriveSpeed.ItemsSource = _driveSpeeds;
-            cmb_DriveSpeed.SelectedItem = 8;
+            var values = UIElements.GetAllowedDriveSpeedsForMediaType(_currentMediaType);
+            cmb_DriveSpeed.ItemsSource = values;
+            cmb_DriveSpeed.SelectedIndex = values.Count / 2;
```
mnadareski: Clarification question: Why use the middle value instead of something like the set value of 8? Is the middle value more preferable?

Data/Constants.cs
```diff
@@ -1,13 +1,45 @@
...
+        private static IReadOnlyList<int> AllowedDriveSpeedsForCD { get; } = new List<int> { 1, 2, 3, 4, 6, 8, 12, 16, 20, 24, 32, 40, 44, 48, 52, 56, 72 };
+        private static IReadOnlyList<int> AllowedDriveSpeedsForDVD { get; } = AllowedDriveSpeedsForCD.Where(s => s <= 24).ToList();
+        private static IReadOnlyList<int> AllowedDriveSpeedsForBD { get; } = AllowedDriveSpeedsForCD.Where(s => s <= 16).ToList();
+        private static IReadOnlyList<int> AllowedDriveSpeedsForUnknownType { get; } = new List<int> { 1 };
+
+        public static IReadOnlyList<int> GetAllowedDriveSpeedsForMediaType(MediaType? type)
+        {
+            switch (type)
+            {
+                case MediaType.CD:
```
mnadareski: CD and GD-ROM both share these drive speeds

Data/Constants.cs
```diff
@@ -1,13 +1,45 @@
...
+                    return AllowedDriveSpeedsForCD;
+                case MediaType.DVD:
```
mnadareski: DVD, HD-DVD, GC, Wii, and WiiU all (in theory) share these drive speeds

Data/Constants.cs
```diff
@@ -1,13 +1,45 @@
...
+                    return AllowedDriveSpeedsForDVD;
+                //TODO: we return them all since DIC doens't support them in any case
+                case MediaType.BluRay:
+                    return AllowedDriveSpeedsForCD;
```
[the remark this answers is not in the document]
mnadareski: I just saw your comment, while that TODO is true, we mind's well be semantically correct where possible

Data/Constants.cs
```diff
@@ -1,13 +1,45 @@
...
+                default:
+                    return AllowedDriveSpeedsForCD;
```
mnadareski: Did you mean to use `AllowedDriveSpeedsForUnknownType` here?

Data/Constants.cs
Jakz: Yes I didn't know it so I left them to be filled by you (or by me since you liste them all)

Data/Constants.cs
Jakz: Same as below

MainWindow.xaml.cs
Jakz: Because I wasn't sure we can assume that 8 is always present, we should decide the best fit here.

MainWindow.xaml.cs
Jakz: Because `OptionsWindow` calls `MainWindow.OnOptionsUpdated` so that we can adjust things that should change (eg `GetOutputNames()`)

OptionsWindow.xaml.cs
Jakz: Because my OCD is suffering, having an lowercase with underscore method name when everything is `CamelCased` really hurt me, can't we find a good trade of like `OnButtonAcceptClick`? `OnComboBoxDriveSpeedChange`? Just throwing ideas

MainWindow.xaml.cs
Jakz: I commented it to avoid slow downs but I didn't push the commit to restore it!

OptionsWindow.xaml.cs
mnadareski: If we do this, I'd prefer if we standardized across all the callbacks. But that's my own OCD :P

mnadareski:
Looks good!

PR #84, opened 2018-07-04
Refactor of ComboBox underlying types

Jakz:
This PR is aimed to solve how ComboBox were populated. The old approach used tuples to just wrap together values but without any semantic. 

This has been refactored:

- added `KnownSystemCategory` enum
- added `KnownSystem` special marker values to split `KnownSystems` into categories
- added `KnownSystemComboBoxItem` which can hold a `KnownSystem` or a `KnownSystemCategory`
- rewrote system population function (and split the actual population for the UI only, with headers)
- adjusted `ComboBox` bindings to use this new approach
- added `Drive` class to store letter, volume label and `isFloppy` flag
- adjusted drives `ComboBox` to use this new object
- modified `DumpEnvironment` to use `Drive` instead that separated attributes
- removed `OrderedDictionary` (not needed anymore and hopefully we won't need it again)

DICUI is now 100% `KeyValuePair`/`Tuple` free! :)


DICUI.Test/Utilities/DriveTest.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace DICUI.Test.Utilities
+{
+    public class DriveTest
+    {
+        [Fact]
+        public void DriveConstructorsTest()
+        {
+            Assert.True(Drive.Floppy('a').IsFloppy);
+            Assert.False(Drive.Optical('d', "test").IsFloppy);
```
mnadareski: Only issue here is that on each unique computer that these tests run on, these tests may or may not pass. For example, on mine, I don't have A connected at all times, so it would fail when this test runs.

DICUI/Data/Enumerations.cs
```diff
@@ -127,11 +133,25 @@ public enum KnownSystem
         TomyKissSite,
         VideoCD,
 
+        MarkerOtherEnd,
+
         #endregion
 
-        Custom = -1
+        Custom = 0x0EADBEEF
```
mnadareski: What is this value now? It seems pretty arbitrary.

DICUI/MainWindow.xaml.cs
```diff
@@ -9,15 +9,16 @@
 using WinForms = System.Windows.Forms;
 using DICUI.Data;
 using DICUI.Utilities;
+using static DICUI.Data.UIElements;
```
mnadareski: Didn't know that `static` was a valid identifier for using statements

DICUI.Test/Utilities/DriveTest.cs
Jakz: This test is not for an effective drive but just for the construction of the `Drive` object, nothing from the machine is actually used.

DICUI/Data/Enumerations.cs
Jakz: I'll change it to something less arbitrary, it was supposed to be `0xDEADBEEF` but since it's signed I couldn't use it.

DICUI/MainWindow.xaml.cs
```diff
@@ -226,8 +246,6 @@ private void BrowseFolder()
         private DumpEnvironment DetermineEnvironment()
         {
             // Populate all KVPs
```
mnadareski: Just remove this comment as well since it's no longer relevant

DICUI/Tasks.cs
```diff
@@ -49,15 +49,16 @@ public class DumpEnvironment
...
+        public Drive Drive;
         public KnownSystem? System;
         public MediaType? Type;
-        public bool IsFloppy;
         public string DICParameters;
 
         // External process information
         public Process dicProcess;
 
+        public bool IsFloppy { get => Drive.IsFloppy; }
```
mnadareski: you can keep this in the same location where `public bool IsFloppy;` was, it looks cleaner that way

DICUI/Utilities/Validators.cs
```diff
@@ -378,54 +396,27 @@ public static class Validators
...
-                        break;
-                }
-
-                systemsDict.Add(Converters.KnownSystemToString(system), system);
-            }
-
-            return systemsDict;
+            return Enum.GetValues(typeof(KnownSystem))
+                .OfType<KnownSystem?>()
+                .Where(s => !s.IsMarker())
```
mnadareski: Just a double check: this doesn't remove the markers from the combo box?

DICUI.Test/Utilities/DriveTest.cs
mnadareski: Oh okay, just making sure.

mnadareski:
Looks good, all minor stuff can be cleaned up in a future commit.

PR #90, opened 2018-07-09
Added new options

Jakz:
This PR adds some customization to the DIC process, it adds 4 options which can be changed in specific window.

- Fixed `Unknown` system which wasn't present anymore in combobox
- Created new `ViewModels` file which should contain all the view models needed by the UI
- Added Quiet mode option to disable DIC sounds
- Added Paranoid mode option which should enable all pedantic flags
- Added Disable Media Type Detection to disable detection at startup (which is not always needed and can slow down startup by a lot
- Added configurable /c2 flag reread amount value

This PR is also a proof of concept about how binding between data and views should be done in WPF (so that we should follow this approach in general).

I left 2 TODOs, both relative to `KnownSystemAndMediaTypeToParameters` since paranoid flags and their tests should be handled by someone with a better knowledge of flag meanings (I choose you @mnadareski!)


DICUI/MainWindow.xaml.cs
```diff
@@ -228,6 +235,8 @@ private void PopulateSystems()
 
             List<KnownSystemComboBoxItem> comboBoxItems = new List<KnownSystemComboBoxItem>();
 
+            comboBoxItems.Add(new KnownSystemComboBoxItem(KnownSystem.NONE));
```
mnadareski: Check for an "Unknown" entry under the first group as well, it seems like it might have been shifted to there originally, hence I thought it was "missing".

DICUI/Utilities/Converters.cs
```diff
@@ -271,20 +271,26 @@ public static List<string> KnownSystemAndMediaTypeToParameters(KnownSystem? sys,
...
                     {
                         case KnownSystem.AppleMacintosh:
                         case KnownSystem.IBMPCCompatible:
                             parameters.Add(DICFlags.NoFixSubQSecuROM);
-                            parameters.Add(DICFlags.ScanFileProtect);
-                            parameters.Add(DICFlags.ScanSectorProtect);
+
+                            if (paranoid)
+                            {
+                                parameters.Add(DICFlags.ScanFileProtect);
```
mnadareski: I'd claim that `ScanFileProtect` is something that should be done regardless.

DICUI/Utilities/Converters.cs
```diff
@@ -262,7 +262,7 @@ public static string KnownSystemAndMediaTypeToBaseCommand(KnownSystem? sys, Medi
         /// <param name="sys">KnownSystem value to check</param>
         /// <param name="type">MediaType value to check</param>
         /// <returns>List of strings representing the parameters</returns>
-        public static List<string> KnownSystemAndMediaTypeToParameters(KnownSystem? sys, MediaType? type)
+        public static List<string> KnownSystemAndMediaTypeToParameters(KnownSystem? sys, MediaType? type, bool paranoid, uint RereadAmountC2)
```
mnadareski: Curious nit: Why `uint` for the reread amount? Also, might want to have a case where `-1` means remove the count entirely, since a blank count defaults to something like 4000 rereads

DICUI/Utilities/Converters.cs
```diff
@@ -271,20 +271,26 @@ public static List<string> KnownSystemAndMediaTypeToParameters(KnownSystem? sys,
...
 
+            //TODO: adjust parameters according to paranoid mode
+
             // Now sort based on disc type
             List<string> parameters = new List<string>();
             switch (type)
             {
                 case MediaType.CD:
-                    parameters.Add(DICFlags.C2Opcode); parameters.Add("20");
+                    parameters.Add(DICFlags.C2Opcode); parameters.Add(RereadAmountC2 > 0 ? Convert.ToString(RereadAmountC2) : "20");
```
mnadareski: If you can have a "paranoid" case for all CDs as well, I'd like to add `DICFlags.SubchannelReadLevel, "2"`, which is the slower but more accurate subchannel read-ahead.

DICUI/Utilities/Converters.cs
```diff
@@ -271,20 +271,26 @@ public static List<string> KnownSystemAndMediaTypeToParameters(KnownSystem? sys,
                 return null;
             }
 
+            //TODO: adjust parameters according to paranoid mode
```
mnadareski: Add a paranoid one for DVD: `DICFlags.CMI`. That's the "Copyright Management Information" flag, and it is highly paranoid for DVDs. This counts for HD-DVD as well, since that's lumped under the DVD read flag.

DICUI/Utilities/Converters.cs
Jakz: Because I don't like to use special values to represent special cases, it's too much ancient C way of doing things. An amount of retries can't be negative, since it has no sense being negative so I used an uint but no problem in any case.

DICUI/MainWindow.xaml.cs
Jakz: Yes it's not a real system so it shouldn't be returned from the list but added manually. This is fixed now.

PR #155, opened 2019-08-25
Sony PlayStation date extraction fixed, 3 separate issues:

superg:
1. 190x year instead of 200x
2. Failure to get date for some titles where SYSTEM.CNF BOOT record doesn't have backslash after "cdrom:"
3. Date shift by one day due to Utc offset applied
Relaxed directory path requirements to allow dot and ampersand. DiscImageCreator had to be changed in order to support this, will be submitted separately (https://github.com/saramibreak/DiscImageCreator/pull/33).


superg:
I will split it to two separate pull requests

PR #286, opened 2021-05-30
Fix for high cpu usage, issue with loop.

BadAd84:
When doing nothing, MPF maxes out a CPU core due to a loop. This wasnt in 1.18, so its fairly recent addition. I have just added a small sleep, but you may want to do it another way.

Either way, fix stops the loop maxing out a core.

### DIFF
--- a/MPF.Library/Data/ProcessingQueue.cs
+++ b/MPF.Library/Data/ProcessingQueue.cs
@@ -61,7 +61,7 @@ namespace MPF.Data
                 {
                     if (this.TokenSource.IsCancellationRequested)
                         break;
-
+                    System.Threading.Thread.Sleep(10);
                     continue;
                 }
 


MPF.Library/Data/ProcessingQueue.cs
```diff
@@ -61,7 +61,7 @@ private void ProcessQueue()
                 {
                     if (this.TokenSource.IsCancellationRequested)
                         break;
-
+                    System.Threading.Thread.Sleep(10);
```
mnadareski, line 64: You can up this number if you want. Half a second is probably reasonable if there was nothing going on. Honestly should have implemented this the first time around.

BadAd84: I can increase if you like, but the value used seems to fix the issue fine for me. CPU usage goes down to 0 and task manager reports low power usage (down from high).

Let me know if you want it at 500ms and I can edit.

PR #292, opened 2021-07-12
Fixed launching MPF in Windows 7.

IcySon55:
Just had to "downgrade" the Aero reference and it now launches and runs just fine in Windows 7.


mnadareski:
It makes me annoyed that this was how Win7 UI reliability things could be fixed. Thank you for finding this and testing it out. Confirmed that the UI still works as expected on Win10.

PR #352, opened 2022-01-31
Add updater using Onova

Zopolis4:
Got tired of updating manually, so I added an updater using [Onova](https://github.com/Tyrrrz/Onova).


JohnVeness:
Personally I'd like to see MPF available in "app stores" like Chocolatey, Scoop etc., rather than self-updating.

Zopolis4:
That's a matter for the package managers to handle, and not one that precludes the implementation of update support in MPF itself.

PR #371, opened 2022-03-04
Fix Update Label button

IcySon55:
System type, media type, and drive speed dropdowns are no longer reset by the Update Label button.

### DIFF
--- a/MPF/ViewModels/MainViewModel.cs
+++ b/MPF/ViewModels/MainViewModel.cs
@@ -478,7 +478,8 @@ namespace MPF.GUI.ViewModels
         /// </summary>
         /// <param name="removeEventHandlers">Whether event handlers need to be removed first</param>
         /// <param name="rescanDrives">Whether drives should be rescanned or not</param>
-        public async void InitializeUIValues(bool removeEventHandlers, bool rescanDrives, bool skipTypeDetection = false)
+        /// <param name="refreshOnly">Whether the initialization only updates the output paths</param>
+        public async void InitializeUIValues(bool removeEventHandlers, bool rescanDrives, bool refreshOnly = false)
         {
             // Disable the dumping button
             App.Instance.StartStopButton.IsEnabled = false;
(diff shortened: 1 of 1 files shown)


MPF/ViewModels/MainViewModel.cs
```diff
@@ -478,7 +478,7 @@ public void UpdateOptions(OptionsWindow optionsWindow)
         /// </summary>
         /// <param name="removeEventHandlers">Whether event handlers need to be removed first</param>
         /// <param name="rescanDrives">Whether drives should be rescanned or not</param>
-        public async void InitializeUIValues(bool removeEventHandlers, bool rescanDrives)
+        public async void InitializeUIValues(bool removeEventHandlers, bool rescanDrives, bool skipTypeDetection = false)
```
mnadareski, line 481: New param needs to be added to the param list in comments.

MPF/ViewModels/MainViewModel.cs
```diff
@@ -516,13 +516,18 @@ public async void InitializeUIValues(bool removeEventHandlers, bool rescanDrives
...
+            {
+                // Determine current media type, if possible
+                await App.Instance.Dispatcher.InvokeAsync(PopulateMediaType);
+                CacheCurrentDiscType();
+                SetCurrentDiscType();
+            }
 
             // Set the initial environment and UI values
-            SetSupportedDriveSpeed();
+            if (!skipTypeDetection)
```
mnadareski, line 529: This `if` can be merged into the proceeding one. Though at second look I don't fully understand why you're gating this off behind the `if` to begin with.

IcySon55: Perhaps it doesn't need to be gated. I will make sure that keeping it doesn't reset the drive speed dropdown.

mnadareski: I think it's reasonable if the speed changes back to the default. This can guard against too-fast or too-slow dumps if the last disc was particularly difficult.

IcySon55:
Superseded by #459 .

PR #378, opened 2022-03-31
Fix RingCodeGuideWindow.xaml so text matches images

JohnVeness:


mnadareski (approved):
Thank you for catching these! Minor, but it helps new users.

Issue #420, opened 2022-10-21
[Request] PS4: parse internal serial from bd/param.sfo


tjanas:
**Is your feature request related to a problem? Please describe.**
Version is parsed and populated from param.sfo. I would like the same for Internal Serial.

**Describe the solution you'd like**
Populate [T:ISN] field in Comments from the value contained in param.sfo.

**Describe alternatives you've considered**
Manually opening the file in a hex editor and copy/pasting the value into my submissionInfo file after it is created.

**Additional context**
![image](https://user-images.githubusercontent.com/20304619/197226889-274f7285-be5e-45f6-8b0b-ccc69e39257e.png)

PR #421, opened 2022-10-21
Populate internal serial from PS4 disc

tjanas:
I haven't compiled or tested this. I followed the pattern for parsing the version.
If it works, it should resolve https://github.com/SabreTools/MPF/issues/420.

### DIFF
--- a/MPF.Modules/BaseParameters.cs
+++ b/MPF.Modules/BaseParameters.cs
@@ -1323,6 +1323,43 @@ namespace MPF.Modules
         }
 
         /// <summary>
+        /// Get the internal serial from a PlayStation 4 disc, if possible
+        /// </summary>
+        /// <param name="driveLetter">Drive letter to use to check</param>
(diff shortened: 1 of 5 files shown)


MPF.Modules/BaseParameters.cs
```diff
@@ -1396,6 +1396,43 @@ protected static string GetPlayStation5Version(char? driveLetter)
             }
         }
 
+        /// <summary>
+        /// Get the internal serial from a PlayStation 4 disc, if possible
+        /// </summary>
+        /// <param name="driveLetter">Drive letter to use to check</param>
+        /// <returns>Internal disc serial if possible, null on error</returns>
+        protected static string GetPlayStation4Serial(char? driveLetter)
```
mnadareski, line 1404: Minor nit: please move this alphabetically. In this case after `GetPlayStation4Version` and before `GetPlayStation5Version`.

mnadareski (approved):

PR #423, opened 2022-10-23
Populate internal serial from PS5 disc

tjanas:
Similar to the extraction of version from param.json
image

### DIFF
--- a/MPF.Modules/BaseParameters.cs
+++ b/MPF.Modules/BaseParameters.cs
@@ -1397,6 +1397,43 @@ namespace MPF.Modules
         }
 
         /// <summary>
+        /// Get the internal serial from a PlayStation 5 disc, if possible
+        /// </summary>
+        /// <param name="driveLetter">Drive letter to use to check</param>
+        /// <returns>Internal disc serial if possible, null on error</returns>
+        protected static string GetPlayStation5Serial(char? driveLetter)
+        {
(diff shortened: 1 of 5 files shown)


MPF.Modules/BaseParameters.cs
```diff
@@ -1396,6 +1396,43 @@ protected static string GetPlayStation4Version(char? driveLetter)
...
+            string paramJsonPath = Path.Combine(drivePath, "bd", "param.json");
+            if (!File.Exists(paramJsonPath))
+                return null;
+
+            // Let's try reading param.json to find the serial in the unencrypted JSON
+            try
+            {
+                using (BinaryReader br = new BinaryReader(File.OpenRead(paramJsonPath)))
+                {
+                    br.BaseStream.Seek(0x82E, SeekOrigin.Begin);
```
mnadareski, line 1425: I'm curious: is it possible to parse out the JSON and then read it _as_ JSON? Or is it partially encrypted?

Whovian9369: Partially encrypted, but (at least in the one I glanced at from the Redump forums) more so at the beginning of the file.

mnadareski: Good enough for now.

mnadareski (approved):

PR #442, opened 2022-12-24
Disable log files compression default in MPF.Check

fuzz6001:
Since the current default is `true`, there is no way to disable compression.


fuzz6001:
Considering users using without "--zip", it may be better to add a no-compression option.

mnadareski:
This is an issue in initialization, not flag setting.

PR #459, opened 2023-02-27
Originally intended behaviour of the Update Label button.

IcySon55:
Sets up the environment and simply updates the output path and filename. No System, Media, Drive, or Speed changes. It's a very fast update meant for large lots of similar discs.

### DIFF
--- a/MPF/ViewModels/MainViewModel.cs
+++ b/MPF/ViewModels/MainViewModel.cs
@@ -573,11 +573,41 @@ namespace MPF.UI.ViewModels
         /// <summary>
         /// Performs a fast update of the output path while skipping disc checks
         /// </summary>
-        private void FastUpdateLabel()
+        /// <param name="removeEventHandlers">Whether event handlers need to be removed first</param>
+        private void FastUpdateLabel(bool removeEventHandlers)
         {
+            // Disable the dumping button
+            App.Instance.StartStopButton.IsEnabled = false;
+
+            // Safely uncheck the parameters box, just in case
+            if (App.Instance.EnableParametersCheckBox.IsChecked == true)
+            {
+                App.Instance.EnableParametersCheckBox.Checked -= EnableParametersCheckBoxClick;
+                App.Instance.EnableParametersCheckBox.IsChecked = false;
+                App.Instance.ParametersTextBox.IsEnabled = false;
+                App.Instance.EnableParametersCheckBox.Checked += EnableParametersCheckBoxClick;
+            }
+
+            // Set the UI color scheme according to the options
+            if (App.Options.EnableDarkMode)
+                EnableDarkMode();
+            else
+                DisableDarkMode();
+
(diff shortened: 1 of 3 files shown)


MPF/ViewModels/MainViewModel.cs
```diff
@@ -1532,7 +1549,10 @@ private void SystemTypeComboBoxSelectionChanged(object sender, SelectionChangedE
         private void UpdateVolumeLabelClick(object sender, RoutedEventArgs e)
         {
             if (_canExecuteSelectionChanged)
-                InitializeUIValues(removeEventHandlers: true, rescanDrives: false);
+                if (App.Options.FastUpdateLabel)
+                    FastUpdateLabel();
+                else
+                    InitializeUIValues(removeEventHandlers: true, rescanDrives: false);
```
mnadareski, line 1555: Nitpicky thing - Can you wrap these 4 lines in curly braces? even though it's mostly apparent, I prefer making it obvious that these are all under the first `if`.

MPF/ViewModels/MainViewModel.cs
```diff
@@ -570,6 +570,53 @@ public async void InitializeUIValues(bool removeEventHandlers, bool rescanDrives
...
+            }
+
+            // Set the UI color scheme according to the options
+            if (App.Options.EnableDarkMode)
+                EnableDarkMode();
+            else
+                DisableDarkMode();
+
+            // Force the UI to reload after applying the theme
+            App.Instance.UpdateLayout();
```
mnadareski, line 598: This part is only needed if the options updated. You should be in the clear to remove this.

MPF/ViewModels/MainViewModel.cs
```diff
@@ -570,6 +570,53 @@ public async void InitializeUIValues(bool removeEventHandlers, bool rescanDrives
...
+
+            // Remove event handlers to ensure ordering
+            if (removeEventHandlers)
```
mnadareski, line 601: You can reasonably have this set to always run and remove the parameter. I appreciate keeping it in, though.

MPF/ViewModels/MainViewModel.cs
```diff
@@ -570,6 +570,53 @@ public async void InitializeUIValues(bool removeEventHandlers, bool rescanDrives
...
+                DisableEventHandlers();
+
+            // Refresh the drive info
+            Drive drive = App.Instance.DriveLetterComboBox.SelectedItem as Drive;
+            drive.RefreshDrive();
+
+            // Set the initial environment and UI values
+            Env = DetermineEnvironment();
+            GetOutputNames(true, true);
```
mnadareski, line 610: Setting `driveChanged` only should be sufficient. Adding the second flag shouldn't be needed as `driveChanged` does nothing but determine that the previous path may have already existed.

mnadareski (approved):

PR #484, opened 2023-04-25
Get write offset from redumper 119

fuzz6001:
Remove fast forwarding because `**** MODE: split` has gone away.


fuzz6001:
This PR works with build 151 (and 152).

mnadareski (approved):
Reasonable enough, thanks for the fix!

PR #493, opened 2023-05-27
Update hardware information pull from redumper

fuzz6001:
Change the current fast-forward keyword as it has become unstable.
I checked with build 151 and 152.


mnadareski (approved):
Also reasonable, thanks again!

PR #542, opened 2023-08-30
Fix PlayStation serial code region parsing

Deterous:
- Do not assume S_J_ and S_P_ serials are Japanese
- Fix typo: PABX -> PBPX
- PCXC are Japanese
- PUBX are USA

### DIFF
--- a/MPF.Modules/BaseParameters.cs
+++ b/MPF.Modules/BaseParameters.cs
@@ -1859,19 +1859,19 @@ namespace MPF.Modules
                                 // Check first two digits of S_PS serial
                                 switch (serial.Substring(4, 2))
                                 {
-                                    case '46': return Region.SouthKorea;
-                                    case '56': return Region.SouthKorea;
-                                    case '51': return Region.Asia;
-                                    case '55': return Region.Asia;
+                                    case "46": return Region.SouthKorea;
+                                    case "56": return Region.SouthKorea;
+                                    case "51": return Region.Asia;
+                                    case "55": return Region.Asia;
                                     default: return Region.Japan;
                                 }
                             case 'M':
                                 // Check first three digits of S_PM serial
                                 switch (serial.Substring(4, 3))
                                 {
-                                    case '645': return Region.SouthKorea;
-                                    case '675': return Region.SouthKorea;
-                                    case '885': return Region.SouthKorea;
+                                    case "645": return Region.SouthKorea;
+                                    case "675": return Region.SouthKorea;
+                                    case "885": return Region.SouthKorea;
                                     default: return Region.Japan; // Remaining S_PM serials may be Japan or Asia
                                 }
                             default: return Region.Japan;


mnadareski:
Please note that this is used by both PS1 and PS2. Don't remove things or change existing cases unnecessarily.

MPF.Modules/BaseParameters.cs
```diff
@@ -1849,9 +1849,8 @@ protected static string GetPlayStation5Version(char? driveLetter)
                     case 'A': return Region.Asia;
                     case 'C': return Region.China;
                     case 'E': return Region.Europe;
-                    case 'J': return Region.JapanKorea;
```
mnadareski, line 1852: I disagree with this change. Users are more than capable of selecting the region that their particular disc is from. There's a reason it's both JP/KR as it is now.

MPF.Modules/BaseParameters.cs
```diff
@@ -1849,9 +1849,8 @@ protected static string GetPlayStation5Version(char? driveLetter)
...
                     case 'K': return Region.SouthKorea;
-                    case 'P': return Region.Japan;
+                    case 'P': return null; // P-region discs may be Japan, Asia or Korea
```
mnadareski, line 1853: It sounds like a more accurate thing would be to return `Region.Asia` instead, then.

MPF.Modules/BaseParameters.cs
```diff
@@ -1861,13 +1860,17 @@ protected static string GetPlayStation5Version(char? driveLetter)
                 return Region.Japan;
 
             // Region appears entirely random
-            else if (serial.StartsWith("PABX"))
+            else if (serial.StartsWith("PBPX"))
```
mnadareski, line 1863: Keep both PABX and PBPX please.

MPF.Modules/BaseParameters.cs
mnadareski: Thank you. Solving issues exactly as requested is not always the right course of action :)

sadikyo: Even if it's not always correct, I'd rather SLPM return Japan, then Asia.  It is much more common to be Japan.

Now, if you want to get a bit more specific, I could list out the number ranges that are likely to be certain regions.  But if you don't want to deal with that, that's ok.

MPF.Modules/BaseParameters.cs
```diff
@@ -1849,10 +1849,33 @@ protected static string GetPlayStation5Version(char? driveLetter)
...
+                                }
+                            case 'M':
+                                // Check first three digits of S_PM serial
+                                switch (serial.Substring(4, 3))
+                                {
+                                    case "645": return Region.SouthKorea;
+                                    case "675": return Region.SouthKorea;
+                                    case "885": return Region.SouthKorea;
+                                    default: return Region.Japan; // Remaining S_PM serials may be Japan or Asia
+                                }
```
mnadareski, line 1876: This feels really difficult to keep up to date, I'll be honest. I understand that it's far more accurate, though.

PR #555, opened 2023-09-28
Update README.md

Ragowit:
Typo


mnadareski:
I had to make a separate change so I wrapped this into that. Thanks for pointing it out.

PR #568, opened 2023-10-10
Fix media combobox not updating

Deterous:
The media type combo box was not updating on user selection, due to the detected media type being chosen first.
I have solved this by clearing the detected media type variable once it has been used.

If you want to preserve the variable, then you can fix this a cleaner way, however pressing 'Scan for discs' does still revert the combo box to the detected media, so this change does do as expected.

### DIFF
--- a/MPF.Core/UI/ViewModels/MainViewModel.cs
+++ b/MPF.Core/UI/ViewModels/MainViewModel.cs
@@ -1326,9 +1326,6 @@ namespace MPF.Core.UI.ViewModels
             Result result = Tools.GetSupportStatus(_environment.System, _environment.Type);
             this.Status = result.Message;
 
-            // Set the index for the current disc type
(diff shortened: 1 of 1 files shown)


mnadareski:
It only looks like the selected index is not updating, not necessarily that the internal data isn't. I tried this out without changes and internally the value was correct.

mnadareski:
I believe I found the actual culprit. At `MainViewModel.cs:L1329` the current disc type is still being set, which is what is causing this double assignment. That was necessary back with the old way of updating, but removing just that line also works as it should. Please try that instead and update your PR when you have a chance. The closure above was accidental.

mnadareski:
That shouldn't be needed anymore. It was when the UI and internals were out of sync, unfortunately.

mnadareski (approved):

Issue #514, opened 2023-07-14
[Problem] Able to select Redumper for HD DVD


FoxhackDN:
**Version**
What version are you using? 

- [ ] Stable release (version here)
- [X] WIP release 2.5.1149

**Build**
What runtime version are you using?

- [X] .NET Framework 4.8 running on (Operating System)
- [ ] .NET 6.0 running on (Operating System)

**Describe the issue**
If you select HD DVD as the system, then select redumper as the dumping program, the dump button is grayed out. That's proper behavior.

If you select a different supported dumping tool for the format (DIC or Aaru), the default command line parameters are loaded and the dump button works again. Also proper behavior.

But if you go and select redumper, the command line parameters box is blanked out, AND the dump button works. If you press it, nothing happens, Redumper says "Media and system combination not supported for Redumper". So the only bug here is the button not being grayed out like it's supposed to.

**To Reproduce**
1. Select Redumper
2. Change media to HD DVD Video
3. Change dumping tool from the selection box to DIC or aaru
4. Change it back to Redumper

**Expected behavior**
The button should be grayed out

**Screenshots**

**Additional context**

FoxhackDN:
Updating for 2.7.0, since recent changes have fixed some things:

1. Set the Default System to "No System Selected", restart MPF if needed.
2. You should see there's no default system selected, so the media type is grayed out.
3. Select Redumper
4. Change media to HD DVD Video
5. The dumping box is active, but the parameters box is empty.

If you click on "Start Dumping", MPF will create a folder in the path shown on the output path as normal but no files will be generated, because "Media and system combination not supported for Redumper" (which is the correct behavior.)

This also seems to happen if you have a default system selected in the options and you change to HD DVD Video.

mnadareski:
Should be fixed by https://github.com/SabreTools/MPF/commit/ae326c1d2f9e973db7416ec01c65a7a48bbb2815 and https://github.com/SabreTools/MPF/commit/e8b0b3efaa463b0a726ec354d40b72aa942cc22c

PR #576, opened 2023-10-12
Disable dumping button when Redumper selected with unsupported media type

Deterous:
Naive solution to #514 
~~Does a check for HDDVD+Redumper every time.
This line can be removed if/when redumper supports HD-DVDs.
Happy to close this if a better solution is worked on.~~
Created helper function that checks whether a given program supports a given media.
Only implemented checks for Redumper + HDDVD or Bluray currently.

### DIFF
--- a/MPF.Core/UI/ViewModels/MainViewModel.cs
+++ b/MPF.Core/UI/ViewModels/MainViewModel.cs
@@ -1584,6 +1584,35 @@ namespace MPF.Core.UI.ViewModels
         }
 
         /// <summary>
+        /// Returns False if a given InternalProgram does not support a given MediaType
+        /// </summary>
+        public bool ProgramSupportsMedia(InternalProgram program, MediaType? media)
+        {
+            switch (program)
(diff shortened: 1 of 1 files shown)


MPF.Core/UI/ViewModels/MainViewModel.cs
```diff
@@ -1591,7 +1591,8 @@ private bool ShouldEnableDumpingButton()
             return Drives != null
                 && Drives.Count > 0
                 && this.CurrentSystem != null
-                && this.CurrentMediaType != null;
+                && this.CurrentMediaType != null
+                && (this.CurrentMediaType != MediaType.HDDVD || this.CurrentProgram != InternalProgram.Redumper);
```
mnadareski, line 1595: If you're doing this, it also doesn't support Blu-ray. I'd prefer a helper method instead of hardcoding into this statement.

mnadareski (approved):
Good for temporary code

PR #585, opened 2023-10-19
Set UDF CD threshold at 800MB

Deterous:
Currently, optical media type detection for UDF filesystems is set to DVDs if the size is at least 400MB.

Based on the PC CDs and DVDs in redump, the false positive rate for DVD detection when using a threshold of 800 MB is less than 3%, with an optimal threshold for CD and DVD false detection rate at approx 830MB. However, I think it is best that MPF keep the threshold at 800 MB as CDs greater than 800 MB in size are technically not following the CD-ROM standard.

I have also changed it slightly to set discs with exactly 800 MB as CDs.

Cumulative distributions for PC DVDs (in red) and PC CDs (in blue) with regards to size, where the crossover represents the point with the "lowest false detection rate":
false-positive
crossover

### DIFF
--- a/MPF.Core/Data/Drive.cs
+++ b/MPF.Core/Data/Drive.cs
@@ -265,13 +265,9 @@ namespace MPF.Core.Data
             }
 
             // Handle optical media by size and filesystem
-            if (this.TotalSize >= 0 && this.TotalSize <= 800_000_000 && this.DriveFormat == "CDFS")
+            if (this.TotalSize >= 0 && this.TotalSize <= 800_000_000 && (this.DriveFormat == "CDFS" || this.DriveFormat == "UDF"))
                 return (MediaType.CDROM, null);
-            else if (this.TotalSize >= 0 && this.TotalSize <= 800_000_000 && this.DriveFormat == "UDF")
-                return (MediaType.CDROM, null);
-            else if (this.TotalSize > 800_000_000 && this.TotalSize <= 8_540_000_000 && this.DriveFormat == "CDFS")
-                return (MediaType.DVD, null);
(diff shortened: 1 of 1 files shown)


MPF.Core/Data/Drive.cs
```diff
@@ -265,13 +265,13 @@ public static List<Drive> CreateListOfDrives(bool ignoreFixedDrives)
...
                 return (MediaType.CDROM, null);
-            else if (this.TotalSize >= 0 && this.TotalSize < 400_000_000 && this.DriveFormat == "UDF")
+            else if (this.TotalSize >= 0 && this.TotalSize <= 800_000_000 && this.DriveFormat == "UDF")
                 return (MediaType.CDROM, null);
-            else if (this.TotalSize >= 800_000_000 && this.TotalSize <= 8_540_000_000 && this.DriveFormat == "CDFS")
+            else if (this.TotalSize > 800_000_000 && this.TotalSize <= 8_540_000_000 && this.DriveFormat == "CDFS")
                 return (MediaType.DVD, null);
-            else if (this.TotalSize >= 400_000_000 && this.TotalSize <= 8_540_000_000 && this.DriveFormat == "UDF")
+            else if (this.TotalSize > 800_000_000 && this.TotalSize <= 8_540_000_000 && this.DriveFormat == "UDF")
                 return (MediaType.DVD, null);
```
mnadareski, line 275: All of these checks can be condensed back down to 2 cases since the `this.DriveFormat` checks have been made irrelevant.

mnadareski: Honestly, those are the only two "filesystems" that I've ever seen pop up during testing, so I'm pretty sure that the drive format doesn't get much more granular... unfortunately.

mnadareski (approved):
Sciencey

PR #592, opened 2023-10-31
Remove psxt001z Pkg Ref in MPF.Test

Deterous:
The only `using` for psxt001z inside MPF.Test was deleted but the csproj still has it as a reference.

Note:
It would be nice to put the Libcrypt check from psxt001z into BOS, having the LibCrypt check require its own library doesn't make sense, especially since its only ever used once inside MPF.Core/Protection.cs: https://github.com/SabreTools/MPF/blob/master/MPF.Core/Protection.cs#L142
But I'll leave that to you to decide, I know it's not originally your code.


mnadareski (approved):
Spoilsport

PR #596, opened 2023-11-03
Focus main window after child window closes

Deterous:
Occasionally, the main window hides behind another application after the options window or disc information window is closed.
To replicate current problem:
1. Go to Options
2. Go to the Paths tab, then click on a `...` button so that the windows explorer window pops up
3. Close this window, then close the options window
4. The Main MPF window gets hidden behind previously opened application.

This attempts to solve the problem, but feels like a band-aid solution.
Open to other ideas.


mnadareski (approved):
WPF is lovely

PR #602, opened 2023-11-05
Enable browsing for Redumper path

Deterous:
The options window currently has a non-functional button to browse for the Redumper executable.
This enables it.


mnadareski (approved):
Click

Issue #617, opened 2024-01-02
[Problem] Only in redumper, BD-ROM-66 is reported incorrectly


fuzz6001:
# Version
What version are you using?

- [x] WIP release (3.0.3-1497)

# Build
What runtime version are you using?

- [x] .NET 8.0 running on (Windows 10)

# Describe the issue
Only in redumper, BD-ROM-66 is reported incorrectly.
```
Media Type: BD-ROM-50
```

In DIC, BD-ROM-66 is reported.
```
Media Type: BD-ROM-66
```

Incidentally, the PIC files for both match.

# To Reproduce

1. Dump by redumper
2. Run MPF.Check

# Expected behavior
Output media type BD-ROM-66.

# Additional context
<details>
<summary>redumper !submissionInfo.txt</summary>

```
Common Disc Info:
	Title: (REQUIRED)
	Foreign Title (Non-latin): (OPTIONAL)
	Disc Number / Letter: (OPTIONAL)
	Disc Title: (OPTIONAL)
	System: Sony PlayStation 5
	Media Type: BD-ROM-50
	Category: Games
	Region: SPACE! (CHANGE THIS)
	Languages: SILENCE! (CHANGE THIS)
	Disc Serial: (REQUIRED, IF EXISTS)

	Ringcode Information:

Layer 0 (Outer) Mastering Code (laser branded/etched): (REQUIRED, IF EXISTS)
Layer 0 (Outer) Mastering SID Code: (REQUIRED, IF EXISTS)
Layer 0 (Outer) Toolstamp or Mastering Code (engraved/stamped): (REQUIRED, IF EXISTS)
Data Side Mould SID Code: (REQUIRED, IF EXISTS)
Data Side Additional Mould: (REQUIRED, IF EXISTS)
Layer 1 (Inner) Mastering Code (laser branded/etched): (REQUIRED, IF EXISTS)
Layer 1 (Inner) Mastering SID Code: (REQUIRED, IF EXISTS)
Layer 1 (Inner) Toolstamp or Mastering Code (engraved/stamped): (REQUIRED, IF EXISTS)
Label Side Mould SID Code: (REQUIRED, IF EXISTS)

	Barcode: (OPTIONAL)
	Comments: (OPTIONAL)
	Contents: (OPTIONAL)

Version and Editions:
	Version: 
	Edition/Release: (VERIFY THIS) Original

Extras:
	Primary Volume Descriptor (PVD):

[...]

PR #619, opened 2024-01-03
Get BD PIC Identifier for redumper

Deterous:
Identifies blu-ray PIC for redumper bluray discs.
Is required for small-sized BD-66 discs, to differentiate from BD-50.
Solves #617

### DIFF
--- a/MPF.Core/Modules/Redumper/Parameters.cs
+++ b/MPF.Core/Modules/Redumper/Parameters.cs
@@ -363,6 +363,9 @@ namespace MPF.Core.Modules.Redumper
                         }
 
                         info.Extras!.PIC = GetPIC($"{basePath}.physical", trimLength) ?? string.Empty;
+
+                        var di = InfoTool.GetDiscInformation($"{basePath}.physical");
+                        info.SizeAndChecksums!.PICIdentifier = InfoTool.GetPICIdentifier(di);
                     }
 
                     break;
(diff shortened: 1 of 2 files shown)


MPF.Core/Modules/Redumper/Parameters.cs
```diff
@@ -363,6 +363,9 @@ public override void GenerateSubmissionInfo(SubmissionInfo info, Options options
                         }
 
                         info.Extras!.PIC = GetPIC($"{basePath}.physical", trimLength) ?? string.Empty;
+
+                        var di = InfoTool.GetDiscInformation($"{basePath}.physical");
+                        info.SizeAndChecksums!.PICIdentifier = InfoTool.GetPICIdentifier(di);
```
mnadareski, line 368: I think this can be put before the previous line and then the PIC object that comes out can be passed into the `GetPIC` method instead.

Deterous: `GetDiscInformation` deserializes the PIC, so I think it would need re-serialising before passing to GetPIC, which defeats the purpose.

mnadareski: Alright then, good enough.

mnadareski:
Please add the name of the PR to the changelog as well and I'll approve and merge.

mnadareski (approved):

PR #623, opened 2024-01-09
Use PSX/PS2 serial as filename when Volume Label not present

### LINKED ISSUE #494: [Request] Use PS2 disc ID (e.g. SCES-12345) as folder/ISO name, if no volume name
When dumping PS2 games, if there is no volume name present, the folder and ISO name "track" is used. This is awkward if you are dumping multiple PS2 discs with the same no-volume name.

The disc ID used instead (only in situations where you would otherwise use "track"), i.e. the thing like SCES-12345. I believe this is relatively easily obtained from the files on the disc.

At the moment I rename the "track" folder after each dump, which works, but is slightly annoying.

**Additional context**
Presumably this could apply to other systems too, but it's PS2 discs that I'm working on at the moment.

Deterous:
If no volume label is found, and the current system is PSX or PS2, use the PlayStation serial as the volume label instead of "track"
I initially tried for PS3/PS4/PS5 as well, but those systems have a default volume label ("PS3VOLUME" etc), so there is no point doing this for those systems.

Fixes #494

### DIFF
--- a/MPF.Core/Data/Drive.cs
+++ b/MPF.Core/Data/Drive.cs
@@ -58,6 +58,8 @@ namespace MPF.Core.Data
 
         /// <summary>
         /// Media label as read by Windows, formatted to avoid odd outputs
+        /// If no volume label present, use PSX or PS2 serial if valid
+        /// Otherwise, use "track" as volume label
         /// </summary>
         public string? FormattedVolumeLabel
         {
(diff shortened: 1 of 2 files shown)


MPF.Core/Data/Drive.cs
```diff
@@ -66,10 +68,21 @@ public class Drive
                 string? volumeLabel = Template.DiscNotDetected;
                 if (this.MarkedActive)
                 {
-                    if (string.IsNullOrEmpty(this.VolumeLabel))
-                        volumeLabel = "track";
-                    else
+                    if (!string.IsNullOrEmpty(this.VolumeLabel))
```
mnadareski, line 71: Since this `if` statement is now doing more heavy lifting, please have it so that the `if (this.MarkedActive)` check is inverted and just returns `volumeLabel` immediately.

MPF.Core/Data/Drive.cs
```diff
@@ -66,10 +68,21 @@ public class Drive
...
                         volumeLabel = this.VolumeLabel;
+                    else
+                    {
+                        RedumpSystem? system = this.GetRedumpSystem(null);
+                        if (system == RedumpSystem.SonyPlayStation || system == RedumpSystem.SonyPlayStation2)
```
mnadareski, line 76: If you want to prepare for other possible systems in the future, it would be worthwhile changing this `if`/`else` into a `switch` to make it easier to expand.

mnadareski (approved):

Issue #458, opened 2023-02-27
[Request] System as part of output path


mnadareski:
Add a new option (disabled by default) to include the shortname OR longname of the system in the output path. This may be useful for... someone.

IcySon55:
Tag on Media Type as an optional path component as well for those that care. ^_^ (default off, default after system, can it be toggled to before system? up for discussion.)

mnadareski:
I mentioned this in Discord, but I'd prefer to have a static order to the output, so "SYSTEM/MEDIA" with either being toggleable. Not intending on getting this looked at until after next stable at earliest.

IcySon55:
I may have some other ideas about this as well. ^_^

Deterous:
The user being able to input variables into the Output Path box seems the best approach. The following could be evaluated once dumping has started to determine the output path:
- `%SYSTEM%` (e.g. /PlayStation 2/)
- `%SHORT_SYSTEM%` (e.g. /PS2/)
- `%MEDIA%` (e.g. /CD-ROM/)
- `%PROGRAM%` (e.g. /Redumper/)
- `%DATE%` (e.g. /20240108/)

The output path also contains the filename, which could be set to variables as well:
- `%VOLUME_LABEL%`
- `%SERIAL%` (for supported systems as mentioned in #494 
- `%TITLE%` (If the file gets renamed after the Disc Information Window is closed)

The default output path could then be as complex as e.g. `ISO/%SHORT_SYSTEM%/%MEDIA%/%PROGRAM%/%DATE%`

PR #624, opened 2024-01-10
Allow variables in output path

Deterous:
Attempts to provide a general solution/fix for #458 

Type the following keywords into output path or default output path textboxes to use variables:
`%SYSTEM%` = Long name for system
`%SYS%` = Short name for system
`%MEDIA%` = Long name for media type
`%PROGRAM%` = Long name for dumping program
`%PROG%` = Short name for dumping program (currently only DiscImageCreator -> DIC)
`%LABEL%` = Volume Label
`%DATE%` = "YYYYMMDD" current date
`%DATETIME%` = "YYYYMMDD-HHMMSS" current datetime

For now, these are not documented anywhere, but are provided for advanced users.

### DIFF
--- a/MPF.Core/UI/ViewModels/MainViewModel.cs
+++ b/MPF.Core/UI/ViewModels/MainViewModel.cs
@@ -169,21 +169,21 @@ namespace MPF.Core.UI.ViewModels
         {
             get
             {
-                string system_long = this._currentSystem.LongName() ?? "Unknown System";
-                string system_short = this._currentSystem.ShortName() ?? "unknown";
-                string media_long = this._currentMediaType.LongName() ?? "Unknown Media";
+                string systemLong = this._currentSystem.LongName() ?? "Unknown System";
+                string systemShort = this._currentSystem.ShortName() ?? "unknown";
+                string mediaLong = this._currentMediaType.LongName() ?? "Unknown Media";
                 string program = this._currentProgram.ToString() ?? "Unknown Program";
-                string program_short = program == "DiscImageCreator" ? "DIC" : program;
+                string programShort = program == "DiscImageCreator" ? "DIC" : program;
                 string label = this._currentDrive?.VolumeLabel ?? "track";
                 string date = DateTime.Today.ToString("yyyyMMdd");
                 string datetime = DateTime.Now.ToString("yyyyMMdd-HHmmss");
 
                 return _outputPath
-                    .Replace("<SYSTEM>", system_long)
-                    .Replace("<SYS>", system_short)
-                    .Replace("<MEDIA>", media_long)
+                    .Replace("<SYSTEM>", systemLong)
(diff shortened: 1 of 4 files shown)


Deterous:
I've used < and > as markers as they are invalid windows filename/path characters.
However, `Path.GetInvalidPathChars()` does not currently check against < and > so I have added checks for those in `NormalizeOutputPaths`

MPF.Core/UI/ViewModels/MainViewModel.cs
```diff
@@ -160,6 +161,35 @@ public string OutputPath
...
+                string system_long = this._currentSystem.LongName() ?? "Unknown System";
+                string system_short = this._currentSystem.ShortName() ?? "unknown";
+                string media_long = this._currentMediaType.LongName() ?? "Unknown Media";
+                string program = this._currentProgram.ToString() ?? "Unknown Program";
+                string program_short = program == "DiscImageCreator" ? "DIC" : program;
+                string label = this._currentDrive?.VolumeLabel ?? "track";
+                string date = DateTime.Today.ToString("yyyyMMdd");
+                string datetime = DateTime.Now.ToString("yyyyMMdd-HHmmss");
+
+                return _outputPath
```
mnadareski, line 181: If you can find a reasonable way to add this as hovertext for either the main UI or the default output path in the settings window, that would be appreciated.

Deterous: I will add it to the options window only, as the main window currently has no hover text. Also, variables for the default path will be the main use of this. Also also, users that find it in the options window will figure out they can use it in the main window too.

MPF.Core/UI/ViewModels/MainViewModel.cs
```diff
@@ -160,6 +161,51 @@ public string OutputPath
...
+        /// <summary>
+        /// Currently provided output path, with < > variables evaluated
+        /// Not guaranteed to be a valid path
+        /// </summary>
+        public string OutputPathEvaluated
+        {
+            get
+            {
+                string systemLong = this._currentSystem.LongName() ?? "Unknown System";
+                if (String.IsNullOrEmpty(systemLong))
```
mnadareski, line 173: Please replace all `String.IsNullOrEmpty` with `string.IsNullOrEmpty`

MPF.Core/UI/ViewModels/MainViewModel.cs
```diff
@@ -160,6 +161,51 @@ public string OutputPath
         }
         private string _outputPath;
 
+        /// <summary>
+        /// Currently provided output path, with < > variables evaluated
+        /// Not guaranteed to be a valid path
+        /// </summary>
+        public string OutputPathEvaluated
```
mnadareski, line 168: If this is getting unwieldly, feel free to make this property invoke a method instead of a getter.

PR #627, opened 2024-01-15
Retrieve volume label from logs

### LINKED ISSUE #579: [Request] Enhance Volume Label detection and output in !submissioninfo.txt
At the moment MPF appears to prioritize the ISO9660 volume label to output into !submissioninfo.
My suggestion is to output all present volume labels (ISO, Joliet, HFS, UDF, Rock Ridge, etc)

Example:
[T:VOL]
GAMENAME123 (ISO9660)
GameName123 (Joliet)
GameName12345 (HFS)
GameNameABCD (UDF)
...

Deterous:
- Parses ISO label from Redumper logs
- Parses ISO/Joliet/UDF label from DIC logs
- Adds formatted list of labels from logs to [T:VOL] tag, on top of whatever Windows sees
- Adds a `Dictionary<string, List<string>>? VolumeLabels` field to BaseParameters
- Makes `Drive.GetRedumpSystemFromVolumeLabel` a static method 

Attempts to fix #579 

Perhaps needs more testing, and figuring out where in Aaru logs the volume label is.
Note: With this change, Aaru still outputs "[T:VOL] %label%" exactly as prior functionality.

### DIFF
--- a/MPF.Core/SubmissionInfoTool.cs
+++ b/MPF.Core/SubmissionInfoTool.cs
@@ -682,13 +682,12 @@ namespace MPF.Core
         /// <returns>Formatted string of volume labels and their filesystems</returns>
         private static string? FormatVolumeLabels(string? driveLabel, Dictionary<string, List<string>>? labels)
         {
-
             // Must have at least one label to format
-            if (driveLabel == null && (labels == null || labels.Count <= 0))
+            if (driveLabel == null && (labels == null || labels.Count == 0))
                 return null;
 
             // If no labels given, use drive label
-            if (labels == null || labels.Count <= 0)
+            if (labels == null || labels.Count == 0)
             {
                 // Ignore common volume labels
                 if (Drive.GetRedumpSystemFromVolumeLabel(driveLabel) != null)
@@ -698,7 +697,7 @@ namespace MPF.Core
             }
 
             // If only one label, don't mention fs
-            string firstLabel = labels.FirstOrDefault().Key;
+            string firstLabel = labels.First().Key;
             if (labels.Count == 1 && (firstLabel == driveLabel || driveLabel == null))
             {
                 // Ignore common volume labels
@@ -724,7 +723,7 @@ namespace MPF.Core
             }
 
             // Print each label separated by a comma and a space
-            if (volLabels == null)
+            if (volLabels.Count == 0)
                 return null;
 
             return string.Join(", ", [.. volLabels]);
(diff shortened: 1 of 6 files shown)


MPF.Core/SubmissionInfoTool.cs
```diff
@@ -670,6 +671,65 @@ public async static Task<bool> FillFromRedump(Options options, SubmissionInfo in
...
+        #region Helper Functions
+
+        /// <summary>
+        /// Formats a list of volume labels and their corresponding filesystems
+        /// </summary>
+        /// <param name="labels">Dictionary of volume labels and their filesystems</param>
+        /// <returns>Formatted string of volume labels and their filesystems</returns>
+        private static string? FormatVolumeLabels(string? driveLabel, Dictionary<string, List<string>>? labels)
+        {
+
```
mnadareski, line 685: Nitpicky extra newline here

MPF.Core/SubmissionInfoTool.cs
```diff
@@ -670,6 +671,65 @@ public async static Task<bool> FillFromRedump(Options options, SubmissionInfo in
...
+            // Must have at least one label to format
+            if (driveLabel == null && (labels == null || labels.Count <= 0))
```
mnadareski, line 687: Pedantic nit: count can never be less than 0

MPF.Core/SubmissionInfoTool.cs
```diff
@@ -670,6 +671,65 @@ public async static Task<bool> FillFromRedump(Options options, SubmissionInfo in
...
+            {
+                // Ignore common volume labels
+                if (Drive.GetRedumpSystemFromVolumeLabel(driveLabel) != null)
+                    return null;
+
+                return driveLabel;
+            }
+
+            // If only one label, don't mention fs
+            string firstLabel = labels.FirstOrDefault().Key;
```
mnadareski, line 701: Since you know that the labels contains at least one item at this point, I think you can get away with just `.First`

MPF.Core/SubmissionInfoTool.cs
```diff
@@ -670,6 +671,65 @@ public async static Task<bool> FillFromRedump(Options options, SubmissionInfo in
...
+            // Add remaining labels with their corresponding filesystems
+            foreach (KeyValuePair<string, List<string>> label in labels)
+            {
+                // Ignore common volume labels
+                if (Drive.GetRedumpSystemFromVolumeLabel(label.Key) == null)
+                    volLabels.Add($"{label.Key} ({string.Join(", ", [.. label.Value])})");
+            }
+
+            // Print each label separated by a comma and a space
+            if (volLabels == null)
```
mnadareski, line 727: `volLabels` should never be null since you instantiate it on 712. I think the thing you're looking for is if the count is 0

mnadareski (approved):
Nice

PR #628, opened 2024-01-16
Correct missing space in PVD

fuzz6001:
In rare cases, if there is a space at the end of each line in the PVD output, it will be missing.
screenshot 1705405198

Therefore, remove Trim.
screenshot 1705405411


mnadareski (approved):
Didn't even know that could be an issue

PR #631, opened 2024-01-17
Prevent crashing on invalid parameters

### LINKED ISSUE #630: [Problem] MPF it's crashed when I use "--skip-fill" option
**Version**
Latest WIP version (3.0.3-1507 or something)

**Build**
NET 8.0

**Describe the issue**
I need to use the "--skip-fill" option on one disc to fill incorrect sectors with 0x55 filler, but when I want to use this option, after entering it and unchecking the checkbox, the program crashes

**To Reproduce**
1. Check "Parameters" box
2. Write "--skip-fill=" and enter the sectors you want to skip
3.  Uncheck box
4. Program crash

**Expected behavior**
The program should run after entering the desired option

Deterous:
Replaces a bunch of .Parse() with .TryParse() 

Also probably fixes #630

### DIFF
--- a/MPF.Core/Modules/BaseParameters.cs
+++ b/MPF.Core/Modules/BaseParameters.cs
@@ -642,7 +642,8 @@ namespace MPF.Core.Modules
                 (string value, long factor) = ExtractFactorFromValue(parts[i]);
                 if (sbyte.TryParse(value, out sbyte sByteValue))
                     return (sbyte)(sByteValue * factor);
-                if (value.Length > 2 && value[0] == '0' && (value[1] == 'x' || value[1] == 'X') && sbyte.TryParse(value.Substring(2), NumberStyles.AllowHexSpecifier, CultureInfo.InvariantCulture, out sbyte sByteHexValue))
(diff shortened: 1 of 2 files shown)


MPF.Core/Modules/BaseParameters.cs
```diff
@@ -639,7 +639,10 @@ protected bool ProcessBooleanParameter(List<string> parts, string? shortFlagStri
                 i++;
 
                 (string value, long factor) = ExtractFactorFromValue(parts[i]);
-                return (sbyte)(sbyte.Parse(value) * factor);
+                bool isSByte = sbyte.TryParse(value, out sbyte sByteValue);
+                if (!isSByte)
```
mnadareski, line 643: I hate to do this to you, but merging the `TryParse` line and the `if` line for each of these is something that I'd commonly do. You can also, if you feel like it, you can use the variant that can take the `NumberStyles.AllowHexSpecifier` that may also fix future hex-related issues.

mnadareski (approved):
Good `Try`

PR #634, opened 2024-01-19
Normalize Disc Title in Submission Info

### LINKED ISSUE #365: [Request] Support for Definitive Articles in Titles for Alphabetization
List of (known) Definitive Articles that get placed at the end of a dat title for alphabetization in Redump

A
An
The
Der
Die
Das
De
Los
Las
La
El
Il
Des
Du
Le
Les
L'
I (Italian only)

Note: Watch out for stuff like "Lights, Camera, Action" turning into: "A Lights, Cameraction"

Deterous:
Applies NormalizeDiscTitle to the main title field during SubmissionInfo generation.
Does not operate on Foreign Title as that field does not need alphabetization?

Implements the work done to fix #365 (I assume this is the desired output?)
Probably needs more testing

### DIFF
--- a/CHANGELIST.md
+++ b/CHANGELIST.md
@@ -17,6 +17,7 @@
 - Prevent crashing on invalid parameters (Deterous)
 - Detect CDTV discs (Deterous)
 - Differentiate CD32 from CDTV (Deterous)
+- Normalise Disc Titles in Submission Info (Deterous)
 
 ### 3.0.3 (2023-12-04)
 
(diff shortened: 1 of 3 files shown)


mnadareski:
You are correct, the foreign and alternate titles do not need alphabetization. This only affects the main title. Unfortunately.

mnadareski (approved):
A step in a direction that I wish we didn't need to take

PR #639, opened 2024-01-30
Parse PSX/PS2/KP2 exe date from logs

### LINKED ISSUE #626: [Request] Read executable date from logs, if available
In a long-standing complaint from a subset of moderators, the current MPF code takes executable dates for PS1 and PS2 directly from the executables themselves, adjusting to UTC to account for international dumpers. Previously, dumpers have used a combination of local datetimes (as seen in whatever timezone they are in) or from the logs provided by the dumping programs.

This request covers trying to read those executable dates from the provided logs as a primary source, falling back on the UTC date if either the date in the logs is invalid or the logs do not contain a date at all. This should help some of the confusion and frustration that the original change to UTC brought.

Deterous:
Parses EXE date from DIC and Redumper logs for PSX/PS2/KP2 discs if it exists, falling back to the previous method of MPF reading the date directly if it does not exist in the logs.

Using Redumper with MPF.Check will get the exe date, but DIC with MPF.Check won't because MPF needs to read the system.cnf file from the drive in order to parse exe date from DIC logs.

Fixes #626

### DIFF
--- a/MPF.Core/InfoTool.cs
+++ b/MPF.Core/InfoTool.cs
@@ -503,9 +503,6 @@ namespace MPF.Core
                 .Replace('_', '-')
                 .Replace(".", string.Empty);
 
-            // Some games may have the EXE in a subfolder
-            serial = Path.GetFileName(serial);
-
             // Get the region, if possible
             region = GetPlayStationRegion(exeName);
 
--- a/MPF.Core/InfoTool.cs
+++ b/MPF.Core/InfoTool.cs
@@ -384,6 +384,66 @@ namespace MPF.Core
             return di.Units[0]?.Body?.DiscTypeIdentifier;
         }
 
+        internal static string? GetPlayStationExecutableName(char? driveLetter)
+        {
+            // If there's no drive letter, we can't get exe name
+            if (driveLetter == null)
+                return null;
+
+            // Convert drive letter to drive path
+            string drivePath = driveLetter + ":\\";
+            return GetPlayStationExecutableName(drivePath);
+        }
+
+        internal static string? GetPlayStationExecutableName(string? drivePath)
+        {
+            // If there's no drive path, we can't get exe name
+            if (string.IsNullOrEmpty(drivePath))
+                return null;
+
+            // If the folder no longer exists, we can't get exe name
+            if (!Directory.Exists(drivePath))
+                return null;
+
+            // Get the two paths that we will need to check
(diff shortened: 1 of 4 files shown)


MPF.Core/Modules/DiscImageCreator/Parameters.cs
```diff
@@ -548,6 +548,8 @@ public override void GenerateSubmissionInfo(SubmissionInfo info, Options options
                         info.CommonDiscInfo.Region = info.CommonDiscInfo.Region ?? pythonTwoRegion;
                         info.CommonDiscInfo.EXEDateBuildDate = pythonTwoDate;
                     }
+                    if (info.CommonDiscInfo!.CommentsSpecialFields!.TryGetValue(SiteCode.InternalSerialName, out string? kp2Serial))
```
mnadareski, line 551: Given that you need to do modifications on the serial at each point to get the date, it seems reasonable to split the code that gets the executable name from both the part that gets the serial and the part that gets the date. This will make it so that you don't have to worry about overwriting with the wrong value or in the wrong order. This goes for all 3 in DIC.

Deterous: Split GetExecutableName from GetExecutableInfo

MPF.Core/InfoTool.cs
```diff
@@ -425,54 +485,26 @@ internal static bool GetPlayStationExecutableInfo(string? drivePath, out string?
...
+            // Serial is most of the EXE name normalized
+            serial = exeName
+                .Replace('_', '-')
+                .Replace(".", string.Empty);
 
-            // If neither can be found, we return false
-            if (string.IsNullOrEmpty(exeName))
-                return false;
+            // Some games may have the EXE in a subfolder
+            serial = Path.GetFileName(serial);
```
mnadareski, line 507: I think this is handled in `GetPlayStationExecutableName` now, at least from how I'm reading this.

Deterous: I didn't want to think about all possible cases so left it just-in-case, e.g. splitting at `;` created a new file path somehow. If you're happy with it being removed, I can remove it.

mnadareski: Splitting shouldn't cause an issue with the path. The `;` isn't a known directory separator or anything and if it's reassembled the same way, you're golden.

mnadareski (approved):

PR #642, opened 2024-01-31
Verbose Redumper log by default

Deterous:
Sets the Redumper verbose flag by default.
Redumper logs will not look any different if no errors occuring during dumping.


mnadareski (approved):
For the vocal only

PR #647, opened 2024-02-07
Add a GUI for PS3 IRD Creation

### LINKED ISSUE #323: [Request] IRD file creation for PS3 discs
Currently, creating IRD files for PS3 discs mostly requires a PS3. The program Redump2IRD ([HERE](http://forum.redump.org/topic/14035/various-ps3-tools/)) can make IRD files from a Redump .iso but needs:
1- Encrypted Disc key (D1)
2- Decrypted Disc ID (D2)
3- PIC data

The program dumping the PS3 disc to the Redump .iso format with its companion IRD file. Or give the information needed to create the IRD with Redump2IRD.

1- For Encrypted Disc key (D1): Incorporate them into the program or allow the user to provide them (like [PS3 Disc Dumper](https://github.com/13xforever/ps3-disc-dumper)).

2- For Decrypted Disc ID (D2): Maybe through connection to the Redump database which I believe store it or if this could only be extracted with a PS3, set a default value of zeros or {check next section}.

If we set default value to zero's:
Encrypted Disc ID = Decrypted Disc ID
04EC7666382555DE33944D947DD64257 = 00000000000000000000000000000000 

3- For PIC data: I believe that MPF already extracts the needed PIC data since v2.0 as in the changelog it says "Add PIC layerbreak extraction".

I'm not an expert but in trying to understand the PS3 Disc ID (D2) I unpacked 2758 IRD files (downloaded from the now down jonnysp.bplaced.net) and all of them have the Data2 be one of the following:

Encrypted Disc ID = Decrypted Disc ID (with Redump2IRD) = Regions of games
04F9E7546E1035B2BDC9A7547553D332 = 00000000000000FF0002000000000000 = BD50(??)
508EA396776D0928E58D57469F30D059 = 00000000000000FF0002000100000001 = Asia / Japan *mostly*
2D1B1AAE18366BD5CE988BF81D6012FB = 00000000000000FF0002000300000001 = US *mostly*
19795F9BFBD8E3B8B317EDBF6071E728 = 00000000000000FF0002000400000001 = Europe *mostly*
6539A28E2D4311BCD16A7A675559305E = 00000000000000FF0002000600000001 = Europe *only*
5EBD7719383C4A0ACE8331FA74D7263E = 00000000000000FF0002000900000001 = US *only*
2647567752CA7A71243A1EEAA61F50C1 = 00000000000000FF0002000A00000001 = Europe *only*

It seems there is a limited number of PS3 Decrypted Disc ID (D2) that might relate to regions.

**Additional context**
What I am really interested in is the ability to convert between Redump .iso and JB format of my backups. The process that I use:

*To make a Redump .iso from JB folder:
1- Use [PS3 ISO Rebuilder](https://archive.org/details/ps3-iso-rebuilder-1.0.4.1) to build a .iso (plain header) with the proper IRD file.
2- Use [PS3Dec](www.romhacking.net/utilities/1456) with the proper Disc key to encrypt the resulting .iso from step 1

*To extract a Redump .iso to JB folder:
1- Use [PS3Dec](www.romhacking.net/utilities/1456) with the proper Disc key to decrypt the Redump .iso
2- Use [PS3 ISO Rebuilder](https://archive.org/details/ps3-iso-rebuilder-1.0.4.1) to extract the resulting .iso from step 1

So, an IRD file is needed. Even if the D2 and PIC data were wrong, Redump2IRD would still be able to create an IRD file that allows disc verification/rebuilding from JB folder (with [PS3 ISO Rebuilder](https://archive.org/details/ps3-iso-rebuilder-1.0.4.1)). But this will contribute to the spreading of these "sudo" IRD files that are useful for disc verification/rebuilding but might have random values for D2 and PIC data. That's why I suggested that the Decrypted Disc ID (D2) is to be set to zero's if the exact value is not obtainable through a PC dump. 

A more secure solution would be to use the last 8b of the IRD file in a new way. Instead of CRC and UID ([LINK](https://www.psdevwiki.com/ps3/Bluray_disc#IRD_file)), the last 8b of the IRD file would be encrypted. This encrypted section will have:

1- CRC of the unencrypted part of the IRD file (the part before the last 8b)
2- MPF signature only known to SabreTools.

A closed source "Checker" tool is to be used to verify:

1- The unencrypted part of the IRD has not been corrupted/modified by checking it against the CRC in the encrypted section.
2- The encrypted section of the IRD was not corrupted/modified by checking the MPF signature.
3- The IRD was created from an actual disc using MPF by checking the MPF signature. 

With this, we don't need to start a new format, just reusing the old format. Starting a new file format just for disc verification/rebuilding is just not worth it. Also, I believe that most programs that use IRD files (like PS3 ISO Rebuilder) will still work just fine since they don't use the last 8b in any important way.

Also, with this method we can easily verify that two dumps are the same by just comparing the two IRD files (because there is no UID). Increasing our confidence it's a correct dump (if they are the same).

[HERE](https://github.com/ifcaro/PS3-ISO-Rebuilder) is what seems to be the code for PS3 ISO Rebuilder (might be useful)

Deterous:
Needs further testing, and it currently hangs when creating the UI which isn't great user experience, but figured it was good enough for a review in the meantime.

Partially fixes #323 however further work would be needed to enable IRD creation after dumping a PS3 disc.


CHANGELIST.md
```diff
@@ -22,8 +23,9 @@
 - Prevent crashing on invalid parameters (Deterous)
 - Detect CDTV discs (Deterous)
 - Differentiate CD32 from CDTV (Deterous)
-- Skip warning line in Redumper log
 - Normalise Disc Titles in Submission Info (Deterous)
+- Skip warning line in Redumper log
+- Add a GUI for MPF.Check (Deterous)
```
Deterous, line 28: I merged the changelog badly that PR, I'm not rewriting history

MPF.Core/UI/ViewModels/CreateIRDViewModel.cs
```diff
@@ -0,0 +1,1071 @@
+﻿#if NET6_0_OR_GREATER
+using LibIRD;
```
mnadareski, line 2: If creating IRDs is limited to modern .NET anyway, you can scope the entire file here inside of the `#if NET6_0_OR_GREATER`

MPF.Core/UI/ViewModels/CreateIRDViewModel.cs
```diff
@@ -0,0 +1,1071 @@
...
+        public void ChangeLogPath()
+        {
+            if (string.IsNullOrEmpty(this.LogPath))
+            {
+                // No .getkey.log file provided: Reset Key and PIC sections
+                LogPathNotProvided = true;
+
+                Key = null;
+                //KeyStatus = "Will attempt to pull Encryption Key from redump.org"; // Use this when redump key pulling is implemented
+                _keyStatus = "Cannot create an IRD without a key";
```
mnadareski, line 481: Did you mean to assign to `KeyStatus` here instead? It won't trigger a UI update otherwise.

MPF.Core/UI/ViewModels/MainViewModel.cs
```diff
@@ -76,6 +76,20 @@ public Data.Options Options
 
         #region Properties
 
+        /// <summary>
+        /// Indicates the inability to create IRDs
+        /// </summary>
+        public bool CannotCreateIRD
```
mnadareski, line 82: Is there a reason that you're using a negative state here?

MPF.UI.Core/Windows/CreateIRDWindow.xaml.cs
```diff
@@ -0,0 +1,452 @@
+﻿using MPF.Core.UI.ViewModels;
```
mnadareski, line 1: Nitpick about ordering

MPF.Core/UI/ViewModels/CreateIRDViewModel.cs
```diff
@@ -0,0 +1,1071 @@
+﻿#if NET6_0_OR_GREATER
+using LibIRD;
+# endif
+using MPF.Core.Utilities;
```
mnadareski, line 4: Nitpick about ordering

MPF.Core/UI/ViewModels/CreateIRDViewModel.cs
Deterous: You would think so, but it breaks the .xaml as you cannot have preprocessor conditionals inside XAML
Unless you can figure out how to exclude assemblies.

MPF.Core/UI/ViewModels/MainViewModel.cs
Deterous: It binds to whether the "Create PS3 IRD" button in the tool menu is enabled or not.
New property is easier than trying to do negatives in XAML

MPF.UI.Core/Windows/CreateIRDWindow.xaml.cs
Deterous: Do you prefer third-party libraries follow in-built ones?
I think I just let VS sort them for me

MPF.Core/UI/ViewModels/CreateIRDViewModel.cs
Deterous: Last minute copy paste mistake...thanks

MPF.Core/UI/ViewModels/CreateIRDViewModel.cs
mnadareski: This is the view model, not the XAML. The button in the menu to access this could just be hidden or only enabled for modern dotnet.

MPF.UI.Core/Windows/CreateIRDWindow.xaml.cs
mnadareski: System, everything else, static in that order. VS should be defaulting to that.

MPF.Core/UI/ViewModels/MainViewModel.cs
Deterous: Ah, sorry, it is bound to the tooltip being enabled!
The button is bound to CreateIRDMenuItemEnabled
If there was an easy way to negate CreateIRDMenuItemEnabled in XAML then I would do that

mnadareski: If you just gate the functionality behind net6.0+, you won't need a special tooltip. The button just wouldn't be there.

MPF.Core/UI/ViewModels/CreateIRDViewModel.cs
Deterous: The XAML still complains about the non-existence of the view model, and if I define a dummy view model it complains that the methods are not defined. If there's an easy way to do this I'm not finding it.

mnadareski: That is incredibly frustrating. Okay, we can keep it this way until conditional compilation can be figured out.

mnadareski (approved):

PR #649, opened 2024-02-12
Fix DIC log parsing for SS version

Deterous:
MPF currently parses 
> Version of challenge table: 02

in DIC logs as the SS (security sector) version.

This is actually DIC's way of saying the XGD version (1 = Xbox, 2 = Xbox360): https://github.com/saramibreak/DiscImageCreator/blob/31b4d32082a1ec00cf69e6edb369f982001ac807/DiscImageCreator/execScsiCmdforDVD.cpp#L1959

Instead, the SS version for all Kreon dumps using DIC is 1. Useful SSv2 data is only obtainable by 0800 drives, which are not supported by DIC (or any of MPF's current dumping programs).

### DIFF
--- a/CHANGELIST.md
+++ b/CHANGELIST.md
@@ -2,6 +2,7 @@
 
 - Remove .NET 6 from auto-builds
 - Make Redumper the default for new users
+- Fix DIC log parsing for SS version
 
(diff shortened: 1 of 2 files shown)


mnadareski (approved):
TIHI

PR #655, opened 2024-02-22
Opt-in automatic IRD creation after PS3 dump

### LINKED ISSUE #323: [Request] IRD file creation for PS3 discs

Deterous:
Adds yet another checkbox in the Dumping options menu to create an IRD file following a PS3 BD-ROM dump.
This is not by default as it can take up to 10min for a large ISO on an HDD.
The IRD creation will occur following the disc information window, it is the last thing to happen after a PS3 dump. This ensures that if a user is not patient and force-closes MPF, then the logs for that dump have already been zipped, etc.

It will also ONLY occur if the Disc Key field was filled in on the Disc Information Window.
Pulling the key from redump is currently not implemented - perhaps this needs to occur by default if a match is found?

Fixes #323

### DIFF
--- a/MPF.Core/InfoTool.cs
+++ b/MPF.Core/InfoTool.cs
@@ -1495,6 +1495,54 @@ namespace MPF.Core
             return files;
         }
 
+#if NET6_0_OR_GREATER
+        /// <summary>
+        /// Create an IRD and write it to the specified output directory with optional filename suffix
+        /// </summary>
+        /// <param name="outputDirectory">Output folder to write to</param>
+        /// <param name="filenameSuffix">Optional suffix to append to the filename</param>
+        /// <param name="outputFilename">Output filename to use as the base path</param>
+        /// <returns>True on success, false on error</returns>
+        public static (bool, string) WriteIRD(string isoPath, string? discKeyString, string? discIDString, string? picString, long? layerbreak, string? crc32)
+        {
+            try
+            {
+                // Output IRD file path
+                string irdPath = Path.ChangeExtension(isoPath, ".ird");
+
+                // Parse disc key from submission info (Required)
+                byte[]? discKey = Tools.ParseHexKey(discKeyString);
+                if (discKey == null)
+                    return (false, "Failed to create IRD: No key provided");
+                // Parse Disc ID from submission info (Optional)
+                byte[]? discID = Tools.ParseDiscID(discIDString);
+                // Parse PIC from submission info (Optional)
+                byte[]? pic = Tools.ParsePIC(picString);
+                // Parse CRC32 strings into ISO hash for Unique ID field (Optional)
+                uint? uid = Tools.ParseCRC32(crc32);
+                // Ensure layerbreak value is valid (Optional)
+                layerbreak = Tools.ParseLayerbreak(layerbreak);
+
+                // Create Redump-style reproducible IRD
+                LibIRD.ReIRD ird = new(isoPath, discKey, layerbreak, uid);
+                if (pic != null)
+                    ird.PIC = pic;
+                if (discID != null && ird.DiscID[15] != 0x00)
+                    ird.DiscID = discID;
+
+                // Write IRD to file
+                ird.Write(irdPath);
+
+                return (true, "IRD created!");
+            }
+            catch (Exception)
+            {
+                // We don't care what the error is
+                return (false, "Failed to create IRD");
+            }
+        }
+#endif
+
         #endregion
 
         #region Normalization
--- a/MPF.Core/InfoTool.cs
+++ b/MPF.Core/InfoTool.cs
@@ -1503,7 +1503,7 @@ namespace MPF.Core
         /// <param name="filenameSuffix">Optional suffix to append to the filename</param>
         /// <param name="outputFilename">Output filename to use as the base path</param>
         /// <returns>True on success, false on error</returns>
-        public static (bool, string) WriteIRD(string isoPath, string? discKeyString, string? discIDString, string? picString, long? layerbreak, string? crc32)
+        public static async Task<(bool, string)> WriteIRD(string isoPath, string? discKeyString, string? discIDString, string? picString, long? layerbreak, string? crc32)
         {
             try
             {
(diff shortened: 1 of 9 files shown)


MPF.Core/Utilities/Tools.cs
```diff
@@ -71,6 +72,35 @@ private static bool EqualAt(this byte[] stack, byte[] needle, int index)
             return true;
         }
 
+        /// <summary>
+        /// Converts a hex string into a byte array
+        /// </summary>
+        /// <param name="hex">Hex string</param>
+        /// <returns>Converted byte array, or null if invalid hex string</returns>
+        public static byte[]? HexStringToByteArray(string? hexString)
```
mnadareski, line 80: I think there's a helper method for this already?

MPF.Core/Utilities/Tools.cs
```diff
@@ -261,5 +291,315 @@ private static (string? tag, string? url) GetRemoteVersionAndUrl()
...
+
+                // Determine whether GetKey was successful
+                string? line;
+                while ((line = sr.ReadLine()) != null && line.Trim().StartsWith("get_dec_key succeeded!") == false) ;
+                if (line == null)
+                    return false;
+
+                // Look for Disc Key in log
+                while ((line = sr.ReadLine()) != null && line.Trim().StartsWith("disc_key = ") == false) ;
+                // If end of file reached, no key found
```
mnadareski, line 333: Hate to ask this, but if you're putting comments between lines, can you have a newline before the comment? It may be worthwhile to combine some of the comments for logical chunks of code, like places that you do `null` or range checks.

MPF.Core/InfoTool.cs
```diff
@@ -1495,6 +1495,54 @@ private static List<string> GetGeneratedFilePaths(string? outputDirectory, strin
...
+            {
+                // Output IRD file path
+                string irdPath = Path.ChangeExtension(isoPath, ".ird");
+
+                // Parse disc key from submission info (Required)
+                byte[]? discKey = Tools.ParseHexKey(discKeyString);
+                if (discKey == null)
+                    return (false, "Failed to create IRD: No key provided");
+                // Parse Disc ID from submission info (Optional)
+                byte[]? discID = Tools.ParseDiscID(discIDString);
```
mnadareski, line 1518: Same comment here with regards to comments and grouping.

MPF.Core/Utilities/Tools.cs
Deterous: Not that I have seen, this is the same function of mine I originally had in CreateIRDViewModel.
There *is* a slower in-built function for this, but its not available for earlier .NET versions.

mnadareski (approved):
Don't ask for whom the IRD creates, it creates for thee

Issue #535, opened 2023-08-14
[Process Enhancement] GitHub Workflows


mnadareski:
Create a GitHub workflow to start transitioning away from requiring AppVeyor only. Possibly go toward a rolling-release model for WIP builds at the very least.

mnadareski:
https://docs.github.com/en/actions/using-workflows

mnadareski:
I'm calling this complete since the templates are there for CI and we can build up to CD. Thanks @Deterous !

PR #657, opened 2024-02-23
Add CI via GitHub Workflows

Deterous:
Implements in GitHub Actions the exact same CI and WIP builds as Appveyor currently does.
Full framework/runtime builds are commented out, but possible.

Implements #535

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
     - name: Build
       run: dotnet publish ${{ matrix.project }}/${{ matrix.project }}.csproj -f ${{ matrix.framework }} -r ${{ matrix.runtime }} -c ${{ matrix.conf == 'Release' && 'Release -p:DebugType=None -p:DebugSymbols=false' || 'Debug'}} --self-contained true --version-suffix ${{ github.sha }} ${{ (startsWith(matrix.framework, 'net5') || startsWith(matrix.framework, 'net6') || startsWith(matrix.framework, 'net7') || startsWith(matrix.framework, 'net8'))&& '-p:PublishSingleFile=true' || ''}}
       
-    - if: ${{ matrix.project == 'MPF' && matrix.framework == 'net8.0-windows' matrix.conf == 'Debug'}}
+    - if: ${{ matrix.project == 'MPF' && matrix.framework == 'net8.0-windows' && matrix.conf == 'Debug'}}
       name: Bundle Redumper
       run: |
         wget https://github.com/superg/redumper/releases/download/build_311/redumper-2024.01.08_build311-win64.zip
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,75 @@
...
+  pull_request:
+    branches: [ "master" ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        project: [MPF, MPF.Check]
```
mnadareski, line 15: At a later point, it may be worthwhile to make separate GHA workflow files to avoid having to do too many excludes.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,75 @@
...
+      with:
+        dotnet-version: 8.0.x
+        
+    - name: Restore dependencies
+      run: dotnet restore
+      
+    - name: Build
+      run: dotnet publish ${{ matrix.project }}/${{ matrix.project }}.csproj -f ${{ matrix.framework }} -r ${{ matrix.runtime }} -c ${{ matrix.conf == 'Release' && 'Release -p:DebugType=None -p:DebugSymbols=false' || 'Debug'}} --self-contained true --version-suffix ${{ github.sha }} ${{ (startsWith(matrix.framework, 'net5') || startsWith(matrix.framework, 'net6') || startsWith(matrix.framework, 'net7') || startsWith(matrix.framework, 'net8'))&& '-p:PublishSingleFile=true' || ''}}
+      
+    - if: ${{ matrix.project == 'MPF' && matrix.framework == 'net8.0-windows' && matrix.conf == 'Debug'}}
```
mnadareski, line 61: Only bundle with Debug builds? I'm not sure I understand that decision.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,75 @@
...
+      name: Bundle Redumper
+      run: |
```
mnadareski, line 63: It's been just long enough for me not to remember: Is there a way to run the wget step only a single time before the builds and then do the unzip as many times as needed? That would make this easier to update in the future with new bundled builds.

.github/workflows/ci.yml
Deterous: It's what appveyor is currently doing. No point bundling Release when the artifacts aren't uploading

mnadareski: Ah, I see. Keeping it on-par for now.

.github/workflows/ci.yml
Deterous: You mean wget once rather than each of the jobs? The jobs run in parallel, they don't share memory. Updating redumper means updating this script

mnadareski: I guess this only factors in if we decide to bundle more than just Redumper again. The archives tend to be small enough not to be an issue. Yeah, discard this comment.

mnadareski (approved):
I like this as a "first draft" of GHA. This will be fun to mess around with

PR #663, opened 2024-02-27
Add PS3 CFW support to MPF.Check

Deterous:
Had a go at adding PS3 CFW support, which requires parsing .getkey.log, .disc.pic, ISO hash and ISO PVD.
Requires a bit more testing

### DIFF
--- a/MPF.Core/Modules/PS3CFW/Parameters.cs
+++ b/MPF.Core/Modules/PS3CFW/Parameters.cs
@@ -1,0 +1,308 @@
+﻿using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using MPF.Core.Converters;
+using MPF.Core.Data;
+using SabreTools.RedumpLib;
+using SabreTools.RedumpLib.Data;
+
+namespace MPF.Core.Modules.PS3CFW
+{
+    /// <summary>
+    /// Represents a generic set of PlayStation 3 Custom Firmware parameters
+    /// </summary>
+    public class Parameters : BaseParameters
+    {
+        #region Metadata
+
+        /// <inheritdoc/>
+        public override InternalProgram InternalProgram => InternalProgram.PS3CFW;
+
+        #endregion
+
+        /// <inheritdoc/>
+        public Parameters(string? parameters) : base(parameters) { }
+
+        /// <inheritdoc/>
+        public Parameters(RedumpSystem? system, MediaType? type, string? drivePath, string filename, int? driveSpeed, Options options)
+            : base(system, type, drivePath, filename, driveSpeed, options)
+        {
+        }
+
+        #region BaseParameters Implementations
+
+        /// <inheritdoc/>
+        public override (bool, List<string>) CheckAllOutputFilesExist(string basePath, bool preCheck)
+        {
+            var missingFiles = new List<string>();
+
+            
+            if (this.Type != MediaType.BluRay || this.System != RedumpSystem.SonyPlayStation3)
+            {
+                missingFiles.Add("Media and system combination not supported for PS3 CFW");
+            }
+            else
+            {
+                string? getKeyBasePath = GetCFWBasePath(basePath);
+                if (!File.Exists($"{getKeyBasePath}.getkey.log"))
+                    missingFiles.Add($"{getKeyBasePath}.getkey.log");
+                if (!File.Exists($"{getKeyBasePath}.disc.pic"))
+                    missingFiles.Add($"{getKeyBasePath}.disc.pic");
+            }
+            return (missingFiles.Count == 0, missingFiles);
+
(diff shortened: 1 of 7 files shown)


MPF.Core/Modules/PS3CFW/Parameters.cs
```diff
@@ -0,0 +1,308 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Get a isobuster-formatted PVD from a PS3 ISO, if possible
+        /// </summary>
+        /// <param name="isoPath">Path to ISO file</param>
+        /// <param name="pvd">Formatted PVD string, otherwise null</param>
+        /// <returns>True if PVD was successfully parsed, otherwise false</returns>
+        private static bool GetPVD(string isoPath, out string? pvd)
```
mnadareski, line 190: This entire method feels like a good fit for `BaseParameters` or even in one of the places like `InfoTool`. It definitely could be used as a backup for other implementations. If it does move, it may be a good thing to split this into getting the PVD bytes and then formatting the PVD. Personal opinion on the second, of course.

MPF.Core/Modules/PS3CFW/Parameters.cs
```diff
@@ -0,0 +1,308 @@
...
+                    info.Extras.PIC = pic;
+                }
+            }
+
+            // Fill in any artifacts that exist, Base64-encoded, if we need to
+            if (includeArtifacts)
+            {
+                info.Artifacts ??= [];
+
+                if (File.Exists(getKeyBasePath + ".bca"))
```
mnadareski, line 112: Wrong `File.Exists` path

MPF.Core/Modules/PS3CFW/Parameters.cs
```diff
@@ -0,0 +1,308 @@
...
+            info.DumpingInfo.DumpingDate = InfoTool.GetFileModifiedDate(getKeyBasePath + ".getkey.log")?.ToString("yyyy-MM-dd HH:mm:ss");
+
+            // TODO: Put info about abnormal PIC info beyond 132 bytes in comments?
+            if (File.Exists(getKeyBasePath + ".disc.pic"))
+                info.Extras!.PIC = GetPIC(getKeyBasePath + ".disc.pic");
+
+            if (Utilities.Tools.ParseGetKeyLog(getKeyBasePath + ".getkey.log", out string? key, out string? id, out string? pic))
+            {
+                if (key != null)
+                    info.Extras!.DiscKey = key.ToUpper();
```
mnadareski, line 97: Prefer `ToUpperInvariant` over `ToUpper`

MPF.Core/Utilities/Tools.cs
```diff
@@ -389,7 +389,7 @@ public static bool ParseGetKeyLog(string? logPath, out byte[]? key, out byte[]?
                     return false;
 
                 // Convert PIC to byte array
-                pic = Tools.HexStringToByteArray(discPICStr.Substring(0, 230));
+                pic = discPICStr + "00000000";
```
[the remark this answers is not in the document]
Deterous: Good question, I should be doing this differently.
3k3y-style PIC, as stored in IRDs, is 115 bytes (230 hex chars) long.
Redump stores 132 bytes (264 chars).
I *was* only reading the first 8 lines (256 chars) and therefore needed to add 8 more zeroes rather than trim it to 230 chars.
However, I should just be reading the 9th line instead.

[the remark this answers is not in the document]
mnadareski: That sounds preferable since it can always be reduced later.

MPF.Core/Modules/PS3CFW/Parameters.cs
```diff
@@ -0,0 +1,308 @@
...
+                            new()
+                            {
+                                Roms =
+                                [
+                                    new Rom { Name = Path.GetFileName(iso), Size = size.ToString(), Crc = crc, Md5 = md5, Sha1 = sha1 },
+                                ]
+                            }
+                        ]
+                    };
+                }
```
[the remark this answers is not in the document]
Deterous: I got this from somewhere in MPF's code..
But yeah I'll collapse it with UIC's example

MPF.Core/Modules/PS3CFW/Parameters.cs
Deterous: In its current form, I'm not sure I trust it being used for anything other than PS3 ISOs. Does PVD *always* start at 0x8320?
A proper base implementation would look for the magic string "CD001" and go from there, rather than hard coding the position. But, this naive implementation works for PS3 :)

mnadareski: Assuming a 2048-sector ISO, I assume so, but sure, for now this can stay as its own single method.

mnadareski (approved):

PR #670, opened 2024-03-07
Enable LibIRD for all .NET frameworks

Deterous:
No longer limited to .NET 6
However, IRDs created with frameworks prior to .NET 6 will not be byte-identical due to the different GZipStream compression level used. The contents should be the same though.


mnadareski (approved):
SUPPORT THE ANCIENT ONES

PR #672, opened 2024-03-07
Fix Check UI deadlock

Deterous:
No longer hangs when performing async tasks in Check UI
net40 is still broken, which is not a change in functionality

### DIFF
--- a/MPF.Core/UI/ViewModels/CheckDumpViewModel.cs
+++ b/MPF.Core/UI/ViewModels/CheckDumpViewModel.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.IO;
 using System.Linq;
+using System.Threading.Tasks;
 using BinaryObjectScanner;
 using MPF.Core.Data;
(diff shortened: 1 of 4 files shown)


MPF.Core/UI/ViewModels/CheckDumpViewModel.cs
```diff
@@ -179,6 +180,20 @@ public bool CheckDumpButtonEnabled
         }
         private bool _checkDumpButtonEnabled;
 
+        /// <summary>
+        /// Indicates the status of the check dump button
```
mnadareski, line 184: Nit: Fix description

MPF.UI.Core/Windows/CheckDumpWindow.xaml.cs
```diff
@@ -278,6 +286,6 @@ public void SystemTypeComboBoxSelectionChanged(object sender, SelectionChangedEv
                 CheckDumpViewModel.ChangeSystem();
         }
 
-        #endregion
+#endregion
```
mnadareski, line 289: Nit: Indentation was removed

PR #685, opened 2024-04-05
Fix CleanRip hash output for Check

Deterous:
- Only generates ClrMamePro datafile if the mediatype is CDROM
- Deals with case when CleanRip log doesn't have hashes
- Bonus: renames the CleanRip file :)

Should the ISO be hashed regardless of presence of hash in the CleanRip log? (to double check)
What should happen if the hash of the ISO is mismatched with the log?


MPF.Core/Modules/CleanRip/Parameters.cs
```diff
@@ -182,6 +184,30 @@ public override List<string> GetLogFilePaths(string basePath)
...
+                            // Warning
+                        }
+                        if (md5 != string.Empty && md5 != isoMD5)
+                        {
+                            // Warning
+                        }
+                        if (sha1 != string.Empty && sha1 != isoSHA1)
+                        {
+                            // Warning
+                        }
```
mnadareski, line 203: I understand why you want to add these here eventually, but are they needed right now?

mnadareski (approved):

PR #691, opened 2024-04-16
Custom non-redump Redumper options

### LINKED ISSUE #690: [Request] Add an option for redumper --drive-sector-order
My Pioneer drive doesn't support C2 error codes, so I have to specify `--drive-sector-order=DATA_SUB` on the command line every time. It'd be much more convenient if there was a dropdown for the sector order alongside the existing BE_CDDA option.

Alternatively, a way to specify misc extra params to always get added on to the command line would also be nice.

Deterous:
Currently, MPF UI options allows for Redumper BE_CDDA mode and GENERIC drive type.
These are somewhat arbitrary (e.g. why not also allow BE mode, or other drive types) and more importantly can misinform users who may set these options without knowing that they are not supported for redump submission.

This PR gates GENERIC drive mode behind a new "Non-Redump Mode", as well as extends read method beyond just BE_CDDA, and adds a new sector order option (Fixes #690 )

Preview:
image

Checking non-redump options shows a warning:
image

The advanced Redumper options can now be set.
image

Unchecking Non-Redump Options resets the three options back to their defaults.

### DIFF
--- a/CHANGELIST.md
+++ b/CHANGELIST.md
@@ -2,6 +2,7 @@
 
 - Fix parameter parsing for `=` symbol (Deterous)
 - Define better default categories (Deterous)
+- Custom non-redump Redumper options (Deterous)
 
 ### 3.1.5 (2024-04-05)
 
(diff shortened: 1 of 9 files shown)


Deterous:
Changed the combo boxes (and textboxes) to a fixed width
image

mnadareski (approved):
Thorough and handles the concerns I had about these options existing

PR #693, opened 2024-04-30
Option for default Redumper leadin retries

### LINKED ISSUE #692: [Request] Add Redumper "--plextor-leadin-retries=xx" option to MPF
When dumping CDs, sometimes the leadin is not read on the 4th or 5th try and when that happens, I need to close redumper and manually add the option "--plextor-leadin-retries=xx" in the Parameters box.

It would be nice if that option was added automatically by MPF instead of me having to manually enter it.

-Proposed solution
Add the "plextor-leadin-retries" option under "Options"->"Programs"->"Redumper"

-Alternate solution
Add a "Parameters" box under "Options"->"Programs"->"Redumper".
MPF will always append those parameters at the end of the arguments it enters to Redumper.
CAUTION. The user will be responsible for entering appropriate arguments, as the wrong ones MAY invalidate the dump.

Deterous:
Adds a new textbox to set default plextor leadin retries, below the default reread option. Defaults to 4.
Fixes #692

### DIFF
--- a/MPF.Core/Modules/Redumper/Parameters.cs
+++ b/MPF.Core/Modules/Redumper/Parameters.cs
@@ -1134,6 +1134,9 @@ namespace MPF.Core.Modules.Redumper
 
             this[FlagStrings.Retries] = true;
             RetriesValue = options.RedumperRereadCount;
+
+            this[FlagStrings.PlextorLeadinRetries] = true;
+            PlextorLeadinRetriesValue = options.RedumperLeadinRetryCount;
         }
 
         /// <inheritdoc/>
--- a/MPF.Core/Modules/Redumper/Parameters.cs
+++ b/MPF.Core/Modules/Redumper/Parameters.cs
@@ -1135,8 +1135,11 @@ namespace MPF.Core.Modules.Redumper
             this[FlagStrings.Retries] = true;
             RetriesValue = options.RedumperRereadCount;
 
-            this[FlagStrings.PlextorLeadinRetries] = true;
-            PlextorLeadinRetriesValue = options.RedumperLeadinRetryCount;
+            if (options.RedumperEnableLeadinRetry)
+            {
+                this[FlagStrings.PlextorLeadinRetries] = true;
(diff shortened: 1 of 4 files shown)


MPF.Core/Modules/Redumper/Parameters.cs
```diff
@@ -1134,6 +1134,9 @@ protected override void SetDefaultParameters(string? drivePath, string filename,
 
             this[FlagStrings.Retries] = true;
             RetriesValue = options.RedumperRereadCount;
+
+            this[FlagStrings.PlextorLeadinRetries] = true;
+            PlextorLeadinRetriesValue = options.RedumperLeadinRetryCount;
```
mnadareski, line 1139: Did you intend on having this option on all the time? It seems like if this is set to 0, it should be disabled?

Deterous: The default value of 4 *should* be identical to omitting the flag, as redumper has 4 retries by default. Of course this is subject to change in redumper code.
I copied the reread tries option for this one.

mnadareski: Thank you. The reason that I think this would be better to have as an enable-only option is that it does clutter up the resulting commandline unnecessarily for most users.

MPF.UI.Core/Windows/OptionsWindow.xaml
```diff
@@ -443,6 +443,18 @@
                                          ToolTip="Specifies how many rereads are attempted on read error"
                                 />
 
+								<CheckBox VerticalAlignment="Center" Content="Set Default Plextor Lead-in Retries"
+                                          IsChecked="{Binding Options.RedumperEnableLeadinRetry}"
+                                          ToolTip="Enable custom lead-in retries for Plextor drives" Margin="0,4"
+                                />
+								<Label/><!-- Empty label for padding -->
```
mnadareski, line 450: Slight whitespace alignment issue

mnadareski (approved):

PR #701, opened 2024-05-16
Add Xbox Backup Creator support to MPF.Check

### LINKED ISSUE #659: [Request] MPF.Check support for XBC dumps
MPF currently does not support 0800 dumps of Xbox and Xbox360 discs using the Xbox Backup Creator program.
This would require extensive work:
- Parsing the XBC logs, which can include many different lines and can also have multiple "runs" in one log.
- For Original Xbox (XGD1) discs, checking the correct error count exists
- Checking the dump was made with all the correct options (clean ISO)
- For XGD1/XGD2, running SS.bin through ss_sector_range and hashing it for submission info (check if before/after hash differs to know whether we have rawSS.bin)
- Hashing DMI and PFI for submission info

Deterous:
Adds XboxBackupCreator as a program for support within MPF.Check
- Looks for and parses XBC logs
- Rewrite of ss_sector_range and ss-angle-fixer to deal with SS.bin
- Gets SS ranges
- Hashes DMI/PFI/SS
- Gets layerbreak from *.dvd file

Fixes #659


MPF.Core/Modules/BaseParameters.cs
```diff
@@ -1209,6 +1209,60 @@ private static string RemoveHexIdentifier(string value)
             }
         }
 
+        /// <summary>
+        /// Get a isobuster-formatted PVD from a PS3 ISO, if possible
```
mnadareski, line 1213: Nitpick here about saying "PS3 ISO". Maybe something like "2048 byte-per-sector image"?

MPF.Core/Modules/BaseParameters.cs
```diff
@@ -1209,6 +1209,60 @@ private static string RemoveHexIdentifier(string value)
...
+        protected static bool GetPVD(string isoPath, out string? pvd)
+        {
+            pvd = null;
+            try
+            {
+                // Get PVD bytes from ISO file
+                var buf = new byte[96];
+                using (FileStream iso = File.OpenRead(isoPath))
+                {
+                    // TODO: Don't hardcode 0x8320
```
mnadareski, line 1227: I know this is a copy-paste of the original code, but how else would this data be found other than hardcoding the sector offset?

MPF.Core/Utilities/Tools.cs
```diff
@@ -642,5 +643,409 @@ public static bool ParseGetKeyLog(string? logPath, out byte[]? key, out byte[]?
...
+                    return false;
+            }
+        }
+
+        /// <summary>
+        /// Determine if a given SS has already been cleaned
+        /// </summary>
+        /// <param name="ss">Byte array of SS sector</param>
+        /// <returns>True if SS is clean, false otherwise</returns>
+        public static bool IsCleanSS(byte[] ss)
```
mnadareski, line 751: Would all of this security sector stuff be better served in a separate library?

MPF.Core/Modules/XboxBackupCreator/Parameters.cs
```diff
@@ -0,0 +1,684 @@
...
+        /// </summary>
+        /// <param name="baseDir">Base directory to search in</param>
+        /// <returns>Log path if found, null otherwise</returns>
+        private static string? GetLogName(string baseDir)
+        {
+            if (IsSuccessfulLog($"{baseDir}Log.txt"))
+                return $"{baseDir}Log.txt";
+
+            if (IsSuccessfulLog($"{baseDir}log.txt"))
+                return $"{baseDir}log.txt";
```
mnadareski, line 262: This may or may not work the way you're expecting. If you can get someone on Linux to test this, that would be good, especially since Windows is case-insensitive by default.

MPF.Core/Modules/XboxBackupCreator/Parameters.cs
```diff
@@ -0,0 +1,684 @@
...
+        /// <param name="readErrors">Read error count if found, -1 otherwise</param>
+        /// <returns>True if sucessful, otherwise false</returns>
+        private bool GetReadErrors(string? log, out long readErrors)
+        {
+            readErrors = -1;
+
+            if (string.IsNullOrEmpty(log) || !File.Exists(log))
+                return false;
+
+            // TODO: Logic when more than one dump is in the logs
```
mnadareski, line 446: I think it's reasonable (for now) to have it only read the first instance.

MPF.Core/Modules/XboxBackupCreator/Parameters.cs
```diff
@@ -0,0 +1,684 @@
...
+            switch (this.Type)
+            {
+                case MediaType.DVD:
+
+                    // Get PVD from ISO (Currently only for Xbox360)
+                    if (this.System != RedumpSystem.MicrosoftXbox360)
+                    {
+                        if (GetPVD(basePath + ".iso", out string? pvd))
+                            info.Extras!.PVD = pvd;
+                    }
```
mnadareski, line 104: If this is only for X360, then you should have a separate switch block for the system, similar to other modules

MPF.Core/Modules/XboxBackupCreator/Parameters.cs
```diff
@@ -0,0 +1,684 @@
...
+                    // Get Xbox 360 Media ID
+                    if (this.System == RedumpSystem.MicrosoftXbox360)
+                    {
+                        string? mediaID = GetMediaID(logPath);
+                        // TODO: Put Media ID in submission info if not null
+                    }
+                    */
+
+                    // Parse DMI.bin
+                    if (this.System == RedumpSystem.MicrosoftXbox)
```
mnadareski, line 138: Same comment about a separate switch statement here.

MPF.Core/Modules/BaseParameters.cs
Deterous: There's a way to do it properly, reading the volume descriptors (until Volume Descriptor Set Terminator is reached, Type=255), and when Type=1 is found you're reading the PVD. More here: https://wiki.osdev.org/ISO_9660#Volume_Descriptors
But it's 0x8320 for all the images I've dealt with so as long as you know a PVD exists when calling this function..

MPF.Core/Utilities/Tools.cs
Deterous: I'll let you refactor these Tools to something that makes sense.. I thought they'd at least be better served outside of the XBC module as they're not specific to XBC logs. Similar to the PS3 functions above them.

MPF.Core/Modules/XboxBackupCreator/Parameters.cs
Deterous: I can remove the second one as I'm enumerating all txt files right after that anyway. I've just noticed it seems common enough that people have `log` instead of `Log`

MPF.Core/Modules/XboxBackupCreator/Parameters.cs
Deterous: I was thinking people would forget to delete the Log, then dump, then copy the log file. The good dump ends up being at the end..

MPF.Core/Utilities/Tools.cs
mnadareski: Fair enough, that can be looked at later

MPF.Core/Modules/XboxBackupCreator/Parameters.cs
mnadareski: I think that keeping both is fine, I just wanted to see what Linux users experience with this, given the difference in case-sensitivity. I've seen this issue crop up before in other programs.

MPF.Core/Modules/XboxBackupCreator/Parameters.cs
mnadareski: I'm guessing you saw some examples of this when researching?

Deterous: Just my own :)

mnadareski: I'll take it for now.

MPF.Core/Modules/XboxBackupCreator/Parameters.cs
Deterous: It was redundant anyway, no harm done

mnadareski (approved):
Onward and upward

PR #712, opened 2024-06-24
Custom theme colors

Deterous:
Extends Purple theme to allow for any custom foreground and background colors.
Remains a hidden feature.


MPF.UI/Themes/CustomTheme.cs
```diff
@@ -0,0 +1,276 @@
...
+                }
+            }
+
+            return true;
+        }
+
+        /// <summary>
+        /// Converts string hex to int hex
+        /// </summary>
+        private static byte HexToHex(string? hex)
```
mnadareski, line 121: Please convert each of the switches to a call to a helper method that gets the value and just have it multiply by 16 for the first one. You can also probably consolidate the switches in the new syntax:
```
return c switch
{ 
    '0' => 0,
    ...
    _ => 0,
};
```

mnadareski (approved):
Can't wait to use `DEDBEF` as a theme color

PR #724, opened 2024-07-13
Fix logic for deducing region from PlayStation ISN

### LINKED ISSUE #722: [Problem] MPF.Check with "-p /mounted/disc/folder" non-functionality in Linux, for PS2 DVDs
**Version**
What version are you using?

- [x] Stable release (version 3.2.0)

**Build**
What runtime version are you using?

- [x] .NET 8.0 running on (Linux)

**Describe the issue**
This is a follow-up from https://github.com/SabreTools/MPF/issues/706#issuecomment-2212521248

I am attempting to use MPF.Check to generate a !submissionInfo.txt file for a PS2 DVD disc previously dumped with Redumper. As noted in that issue, MPF.Check doesn't trust the disc region from the Redumper log, so needs to check the mounted physical disc, but the behaviour of this appears to be different between MPF.Check in Linux vs Windows.

**To Reproduce**

In Linux:
1. Dump a PS2 DVD, e.g. SLES-51252, in Redumper
2. cd into the folder output from Redumper, e.g. /home/john/redump/SLES-51252
3. Mount the disc as e.g. /run/media/john/333d3c3734333d32 
4. Run `MPF.Check dvd ps2 -u redumper -p /run/media/john/333d3c3734333d32 SLES-51252.iso`
5. Look at the created !submissionInfo.txt file

**Expected behavior**
The !submissionInfo.txt file to contain "Region: Europe" (in this case), as it does correctly do in Windows with `MPF.Check dvd ps2 -u redumper -p d: SLES-51252.iso`.

**Observed behavior**
The !submissionInfo.txt file contains "Region: SPACE! (CHANGE THIS)". It does, however, contain correct PS2 ISN, version and EXE date.

**Additional context**
As a further test, I deleted the following section from the Redumper.log file:
```
PS2 [SLES-51252.iso]:
  EXE: SLES_512.52
  EXE date: 2002-09-20
  version: 1.00
  serial: SLES-51252
  region: Europe
```

and ran the MPF.Check test again. My theory being that this should make MPF.Check do its own check for the EXE name, date, version, serial and region (as opposed to doing its own check just for the region previously), rather than taking that info from the Redumper log. Running 
`MPF.Check dvd ps2 -u redumper -p /run/media/john/333d3c3734333d32 SLES-51252.iso` now shows that the version, ISN, exedate etc. are no longer written to the !submissionInfo.txt file at all.

In other words, as far as I can tell, MPF.Check's code to look for PS2 exe details from the mounted disc doesn't seem to work at all in Linux :(. Unless I'm doing something wrong, which is quite possible!

I should add that listing files in /run/media/john/333d3c3734333d32 does indeed show the system.cnf file, the SLES_512.52 file, etc., and the files are readable.

JohnVeness:
An attempt to fix #722. Please note that I am a complete C# amateur and don't know whether this is the best implementation! And I haven't added it for PS1.

I'm a bit confused about line https://github.com/JohnVeness/MPF/blob/8c9b51914850f583cd70576f500673427d9f4a78/MPF.Frontend/Tools/SubmissionGenerator.cs#L837 which doesn't seem to get executed.

### DIFF
--- a/MPF.Frontend/Tools/SubmissionGenerator.cs
+++ b/MPF.Frontend/Tools/SubmissionGenerator.cs
@@ -698,7 +698,7 @@ namespace MPF.Frontend.Tools
                     if (isDiscImageCreator)
                         info.CommonDiscInfo!.EXEDateBuildDate = DiscImageCreator.GetPlayStationEXEDate($"{basePath}_volDesc.txt", drive?.GetPlayStationExecutableName());
 
-                    if (info.CommonDiscInfo!.CommentsSpecialFields!.TryGetValue(SiteCode.InternalSerialName, out string? kp2Exe) || !string.IsNullOrEmpty(kp2Exe))
+                    if (info.CommonDiscInfo!.CommentsSpecialFields!.TryGetValue(SiteCode.InternalSerialName, out string? kp2Exe) && !string.IsNullOrEmpty(kp2Exe))
                         info.CommonDiscInfo.Region = ProcessingTool.GetPlayStationRegion(kp2Exe);
 
                     if (drive?.GetPlayStationExecutableInfo(out var kp2Serial, out Region? kp2Region, out var kp2Date) == true)
@@ -812,7 +812,7 @@ namespace MPF.Frontend.Tools
                     if (isDiscImageCreator)
                         info.CommonDiscInfo!.EXEDateBuildDate = DiscImageCreator.GetPlayStationEXEDate($"{basePath}_volDesc.txt", drive?.GetPlayStationExecutableName(), psx: true);
 
-                    if (info.CommonDiscInfo!.CommentsSpecialFields!.TryGetValue(SiteCode.InternalSerialName, out string? psxExe) || !string.IsNullOrEmpty(psxExe))
+                    if (info.CommonDiscInfo!.CommentsSpecialFields!.TryGetValue(SiteCode.InternalSerialName, out string? psxExe) && !string.IsNullOrEmpty(psxExe))
                         info.CommonDiscInfo.Region = ProcessingTool.GetPlayStationRegion(psxExe);
 
                     if (drive?.GetPlayStationExecutableInfo(out var psxSerial, out Region? psxRegion, out var psxDate) == true)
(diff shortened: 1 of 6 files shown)


Deterous:
> I'm a bit confused about line https://github.com/JohnVeness/MPF/blob/8c9b51914850f583cd70576f500673427d9f4a78/MPF.Frontend/Tools/SubmissionGenerator.cs#L837 which doesn't seem to get executed.

I think you've spotted the issue at hand; this line was copy-pasted from the other setters but the logic is different as the ISN is being *used*, not set. So the fix here is if ISN is set (trygetvalue is true, isnullorempty is false) -> set region with GetPlayStationRegion

JohnVeness:
Aha, thanks for spotting that. I have made several changes and I _think_ it's now good.

Please triple-check my work before merging (or completely ignore this PR and make your own change!) as I really don't know what I'm doing. You'll probably want to squash the commits into one, too.

MPF.Processors/ProcessingTool.cs
```diff
@@ -372,6 +372,91 @@ static int ReadFromArrayBigEndian(byte[]? bytes, int offset)
             };
         }
 
+        /// <summary>
+        /// Determine the region based on the PlayStation serial code
+        /// </summary>
+        /// <param name="serial">PlayStation serial code</param>
+        /// <returns>Region mapped from name, if possible</returns>
+        public static Region? GetPlayStationRegion(string? serial)
```
mnadareski, line 380: Minor nitpick: Please move this above `GetXGDRegion`

MPF.Frontend/Tools/SubmissionGenerator.cs
```diff
@@ -698,8 +698,8 @@ private static bool ProcessSystem(SubmissionInfo info, RedumpSystem? system, Dri
                     if (isDiscImageCreator)
                         info.CommonDiscInfo!.EXEDateBuildDate = DiscImageCreator.GetPlayStationEXEDate($"{basePath}_volDesc.txt", drive?.GetPlayStationExecutableName());
 
-                    if (!info.CommonDiscInfo!.CommentsSpecialFields!.TryGetValue(SiteCode.InternalSerialName, out string? kp2Exe) || string.IsNullOrEmpty(kp2Exe))
-                        info.CommonDiscInfo.Region = Drive.GetPlayStationRegion(kp2Exe);
+                    if (info.CommonDiscInfo!.CommentsSpecialFields!.TryGetValue(SiteCode.InternalSerialName, out string? kp2Exe) || !string.IsNullOrEmpty(kp2Exe))
```
mnadareski, line 701: Each of these `||` need to be `&&` for safety

MPF.Processors/ProcessingTool.cs
JohnVeness: Done. For alphabetical order reasons?

MPF.Frontend/Tools/SubmissionGenerator.cs
JohnVeness: Done.

mnadareski (approved):

PR #725, opened 2024-07-18
Use actual filenames in report

JohnVeness:
Changes, for example, `Writing information to !submissionInfo.txt...` to `Writing information to !submissionInfo_isoname.txt...`.

### DIFF
--- a/MPF.Frontend/DumpEnvironment.cs
+++ b/MPF.Frontend/DumpEnvironment.cs
@@ -498,7 +498,7 @@ namespace MPF.Frontend
             var filenameSuffix = _options.AddFilenameSuffix ? Path.GetFileNameWithoutExtension(outputFilename) : null;
 
             // Write the text output
-            resultProgress?.Report(ResultEventArgs.Success("Writing information to !submissionInfo.txt..."));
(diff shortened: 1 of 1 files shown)


mnadareski:
The code looks fine, but why are you making this change?

JohnVeness:
It just annoyed me it saying it was outputting to !submissionInfo.txt when it wasn't actually doing so in some circumstances!

Alternatively, I'd be OK if the text was more general like `Writing submission information...` or `Writing protection information...` or ``Writing submission information as JSON...` rather than mentioning the actual filenames.

mnadareski:
I don't see any problem with the current way, but if I were given a choice, I'd prefer the more generic language to matching the output name entirely at that point.

JohnVeness:
It's up to you, of course! This was just me scratching an itch.

mnadareski:
Please change it to the more generic language and I will merge it. Just update this PR, no need to close this one and open a new one.

JohnVeness:
Sorry, had a problem with git branches so ended up making a new PR (#727) after all. Still, at least it gives you a choice!

PR #728, opened 2024-07-19
Remove old --protect-file mentions

JohnVeness:
The --protect-file option itself was removed in e0482aad78d3fa4086ae2dba542ab05c7f0ed3d4 but was still referred to in the help text.


mnadareski (approved):

JohnVeness:
Thanks. I'd like to edit https://github.com/SabreTools/MPF/wiki/MPF-Check but I don't think I have permission...

mnadareski:
Removed

PR #747, opened 2024-09-07
Ensure manufacturer files starting from 0 are zipped in redumper DVD …

### LINKED ISSUE #746: [Problem] MPF does not zip [dump].0.manufacturer file
Occurs in current stable and today's WIP, .NET 8.0 running on Windows 11

**Describe the issue**
When dumping [PS2] Sakura Wars: So Long My Love, which has two dual-layer DVDs, redumper outputs two .manufacturer files:

1. [title].0.manufacturer
2. [title].1.manufacturer

After finishing the dump and filling info, the compression of the log files does not include the "0" manufacturer file and leaves the file chilling outside the zip

**To Reproduce**
Dump a dual layer PS2 game. I don't have any non-PS2 dual layer DVDs to test, but based on the code this could be an issue for any DVD media.

This ocureed for the two Sakura Wars discs and a Yakuza 2 disc I tested

**Expected behavior**
all .manufacturer files are zipped

**Additional context**
in MPF.Processors\Redumper.cs under case MediaType.DVD, the .physical file list accounts for 0, 1, and 2 .physical, but for .manufacturer names it only returns 1 and 2. Adding another entry for $"{baseFilename}.0.manufacturer" fixes the issue for my disks

TurnedToast:
Dual-layer ps2 DVDs (at a minimum) produce a foo.0.manufacturer file in addition to foo.1.manufacturer. This change ensures that 0-file is counted among the logs and compressed when "Compress Log Files" is checked

Fixes #746

### DIFF
--- a/MPF.Processors/Redumper.cs
+++ b/MPF.Processors/Redumper.cs
@@ -459,9 +459,6 @@ namespace MPF.Processors
                         new($"{baseFilename}.1.manufacturer", OutputFileFlags.Binary
                             | OutputFileFlags.Zippable,
                             "manufacturer_1"),
-                        new($"{baseFilename}.2.manufacturer", OutputFileFlags.Binary
-                            | OutputFileFlags.Zippable,
-                            "manufacturer_2"),
                         new([$"{baseFilename}.physical", $"{baseFilename}.0.physical"], OutputFileFlags.Required
                             | OutputFileFlags.Binary
                             | OutputFileFlags.Zippable,
@@ -469,9 +466,6 @@ namespace MPF.Processors
                         new($"{baseFilename}.1.physical", OutputFileFlags.Binary
                             | OutputFileFlags.Zippable,
                             "physical_1"),
-                        new($"{baseFilename}.2.physical", OutputFileFlags.Binary
-                            | OutputFileFlags.Zippable,
-                            "physical_2"),
                         new($"{baseFilename}.security", System.IsXGD()
                             ? OutputFileFlags.Required | OutputFileFlags.Binary | OutputFileFlags.Zippable
                             : OutputFileFlags.Binary | OutputFileFlags.Zippable,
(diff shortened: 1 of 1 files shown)


MPF.Processors/Redumper.cs
```diff
@@ -452,8 +452,11 @@ internal override List<OutputFile> GetOutputFiles(string baseFilename)
...
                             DatfileExists),
-                        new([$"{baseFilename}.manufacturer", $"{baseFilename}.1.manufacturer"], OutputFileFlags.Required
+                        new([$"{baseFilename}.manufacturer", $"{baseFilename}.0.manufacturer"], OutputFileFlags.Required
                             | OutputFileFlags.Binary
+                            | OutputFileFlags.Zippable,
+                            "manufacturer_0"),
+                        new($"{baseFilename}.1.manufacturer", OutputFileFlags.Binary
                             | OutputFileFlags.Zippable,
                             "manufacturer_1"),
                         new($"{baseFilename}.2.manufacturer", OutputFileFlags.Binary
```
mnadareski, line 462: If the numbering has changed to be consistent with `0, 1` instead of `1, 2`, then this can be removed.

mnadareski: `.2.physical` can be removed as well for DVD.

mnadareski: `.3.physical` can be added for Blu-ray.

MPF.Processors/Redumper.cs
```diff
@@ -516,6 +513,9 @@ internal override List<OutputFile> GetOutputFiles(string baseFilename)
                         new($"{baseFilename}.2.physical", OutputFileFlags.Binary
                             | OutputFileFlags.Zippable,
                             "physical_2"),
+                            new($"{baseFilename}.3.physical", OutputFileFlags.Binary
```
mnadareski, line 516: Minor nit: De-indent this line by one level please

mnadareski (approved):

PR #748, opened 2024-09-07
Add _drive.txt file to GetOutputFiles for UmdImageCreator

TurnedToast:
Current WIP (not stable) isn't zipping the _drive log file from UmdImageCreator, this PR just adds it back to the list

### DIFF
--- a/MPF.Processors/UmdImageCreator.cs
+++ b/MPF.Processors/UmdImageCreator.cs
@@ -90,6 +90,10 @@ namespace MPF.Processors
                             | OutputFileFlags.Artifact
                             | OutputFileFlags.Zippable,
                             "disc"),
+                        new($"{baseFilename}_drive.txt", OutputFileFlags.Required
+                            | OutputFileFlags.Artifact
+                            | OutputFileFlags.Zippable,
+                            "disc"),
                         new($"{baseFilename}_mainError.txt", OutputFileFlags.Required
                             | OutputFileFlags.Artifact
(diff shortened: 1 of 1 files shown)


MPF.Processors/UmdImageCreator.cs
```diff
@@ -90,6 +90,10 @@ internal override List<OutputFile> GetOutputFiles(string baseFilename)
                             | OutputFileFlags.Artifact
                             | OutputFileFlags.Zippable,
                             "disc"),
+                        new($"{baseFilename}_drive.txt", OutputFileFlags.Required
+                            | OutputFileFlags.Artifact
+                            | OutputFileFlags.Zippable,
+                            "disc"),
```
mnadareski, line 96: Nit: Update `disc` to `drive`

MPF.Processors/UmdImageCreator.cs
```diff
@@ -90,6 +90,10 @@ internal override List<OutputFile> GetOutputFiles(string baseFilename)
                             | OutputFileFlags.Artifact
                             | OutputFileFlags.Zippable,
                             "disc"),
+                        new($"{baseFilename}_drive.txt", OutputFileFlags.Required
```
mnadareski, line 93: Not entirely sure if this is always output in UIC or if it's only that one branch version. To be safe, the Required flag can probably be removed. It can always be added if it's a guaranteed output.

mnadareski (approved):

PR #749, opened 2024-09-26
[WIP] Fix SafeDisc filtering

TheRogueArchivist:
This will need more work, as it is currently incomplete, unoptimized, and untested.

Once finished, will fix the issue of detected SafeDisc versions being completely filtered away, leaving only generic detections.

### DIFF
--- a/MPF.Frontend/Tools/ProtectionTool.cs
+++ b/MPF.Frontend/Tools/ProtectionTool.cs
@@ -304,7 +304,47 @@ namespace MPF.Frontend.Tools
             // SafeDisc
             if (foundProtections.Any(p => p.StartsWith("SafeDisc")))
             {
-                if (foundProtections.Any(p => Regex.IsMatch(p, @"SafeDisc [0-9]\.[0-9]{2}\.[0-9]{3}", RegexOptions.Compiled) && !p.StartsWith("Macrovision Protection File")))
+                // Best case scenario for new SafeDisc 2+: A full SafeDisc version is found in a line starting with "Macrovision Protect Application". All other SafeDisc detections can be safely scrubbed.
+                // TODO: Scrub "Macrovision Protected Application, " from before the SafeDisc version. Also scrub the "SafeDisc SRV Tool APP" portion of the detection when not needed.
+                if (foundProtections.Any(p => Regex.IsMatch(p, @"SafeDisc [0-9]\.[0-9]{2}\.[0-9]{3}", RegexOptions.Compiled) && p.StartsWith("Macrovision Protected Application")))
+                {
+                    foundProtections = foundProtections.Where(p => !p.StartsWith("Macrovision Protection File"))
+                        .Where(p => !p.StartsWith("Macrovision Security Driver"))
+                        .Where(p => p != "SafeDisc")
+                        .Where(p => !p.StartsWith("Macrovision Protected Application, Macrovision Protected Application [Version Expunged]"))
+                        .Where(p => !(Regex.IsMatch(p, @"SafeDisc [0-9]\.[0-9]{2}\.[0-9]{3}-[0-9]\.[0-9]{2}\.[0-9]{3}", RegexOptions.Compiled)))
+                        .Where(p => !(Regex.IsMatch(p, @"SafeDisc [0-9]\.[0-9]{2}\.[0-9]{3}\+", RegexOptions.Compiled)))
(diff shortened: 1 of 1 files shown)


MPF.Frontend/Tools/ProtectionTool.cs
```diff
@@ -304,7 +304,82 @@ public static string SanitizeFoundProtections(IEnumerable<string> foundProtectio
...
+
+                // Best case scenario for new SafeDisc 2+: A full SafeDisc version is found in a line starting with "Macrovision Protect Application". All other SafeDisc detections can be safely scrubbed.
+                // TODO: Scrub "Macrovision Protected Application, " from before the SafeDisc version. Also scrub the "SafeDisc SRV Tool APP" portion of the detection when not needed.
+                if (foundProtections.Any(p => Regex.IsMatch(p, @"SafeDisc [0-9]\.[0-9]{2}\.[0-9]{3}", RegexOptions.Compiled) && p.StartsWith("Macrovision Protected Application")))
+                {
+                    foundProtections = foundProtections.Where(p => !p.StartsWith("Macrovision Protection File"))
+                        .Where(p => !p.StartsWith("Macrovision Security Driver"))
+                        .Where(p => p != "SafeDisc")
+                        .Where(p => !p.StartsWith("Macrovision Protected Application, Macrovision Protected Application [Version Expunged]"))
```
mnadareski, line 316: This case should never get hit in any of the statements

MPF.Frontend/Tools/ProtectionTool.cs
```diff
@@ -304,7 +304,82 @@ public static string SanitizeFoundProtections(IEnumerable<string> foundProtectio
...
+                        .Where(p => !(Regex.IsMatch(p, @"SafeDisc [0-9]\.[0-9]{2}\.[0-9]{3}-[0-9]\.[0-9]{2}\.[0-9]{3}", RegexOptions.Compiled)))
+                        .Where(p => !(Regex.IsMatch(p, @"SafeDisc [0-9]\.[0-9]{2}\.[0-9]{3}\+", RegexOptions.Compiled)))
+                        .Where(p => !(Regex.IsMatch(p, @"SafeDisc [0-9]\.[0-9]{2}\.[0-9]{3}\/4\+", RegexOptions.Compiled)))
```
mnadareski, line 319: Even though `/4+` is currently the most common, if you're already using regex, it may be safer to change this to `\/[0-9]\+`

mnadareski: This filter is also needed after each of the existing `\+` statements. Its own `if` statement is probably warranted as well.

MPF.Frontend/Tools/ProtectionTool.cs
```diff
@@ -304,7 +304,82 @@ public static string SanitizeFoundProtections(IEnumerable<string> foundProtectio
...
+                // Only SafeDisc 2+ is found.
+                else if (foundProtections.Any(p => p == "SafeDisc 2+"))
+                {
+                    foundProtections = foundProtections.Where(p => p != "SafeDisc")
+                        .Where(p => !p.StartsWith("Macrovision Protection File"))
+                        .Where(p => !(Regex.IsMatch(p, @"SafeDisc [0-9]\.[0-9]{2}\.[0-9]{3}\+", RegexOptions.Compiled)));
+                }
+                // TODO: Make sure edge cases, like SafeDisc Lite, are supported properly.
+                // TODO: Organize and re-add the following cases if applicable: 
+                /*
```
mnadareski, line 381: I know this is still called "WIP" but please don't leave in a large block of commented code.

MPF.Frontend/Tools/ProtectionTool.cs
TheRogueArchivist: Although this was fixed in the latest commit to BOS, it hasn't propagated to MPF yet. So for now, this case can get hit.

MPF.Frontend/Tools/ProtectionTool.cs
TheRogueArchivist: > Even though `/4+` is currently the most common, if you're already using regex, it may be safer to change this to `\/[0-9]\+`

There are currently no other numbers that get reported as`/X+` for SafeDisc. I could add it if you insist, but at the moment it wouldn't make any difference except for making the code very slightly less easy to understand, IMO

TheRogueArchivist: > This filter is also needed after each of the existing `\+` statements. Its own `if` statement is probably warranted as well.

Can you clarify exactly what filter you mean is needed?

mnadareski (approved):
I don't like it, but I also don't like the filtering at all

PR #760, opened 2024-11-19
Fix PIC parser

Deterous:
Currently copy-pasting the PIC from redump.org does not work as the parser only removes the \n character
This fixes it to remove both \n and \r


mnadareski (approved):
Annoying, but makes sense

PR #762, opened 2024-11-22
Improve parameters for default output path

Deterous:
%SYS% now defaults to uppercase (i.e. /DVD-VIDEO/), unless you specify %sys%

The tooltip also works on the label in the options now.

### DIFF
--- a/MPF.Frontend/ViewModels/MainViewModel.cs
+++ b/MPF.Frontend/ViewModels/MainViewModel.cs
@@ -1372,7 +1372,7 @@ namespace MPF.Frontend.ViewModels
 
             return outputPath
                 .Replace("%SYSTEM%", systemLong)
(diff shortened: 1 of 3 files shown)


MPF.Frontend/ViewModels/MainViewModel.cs
```diff
@@ -1372,7 +1372,8 @@ public string EvaluateOutputPath(string outputPath)
 
             return outputPath
                 .Replace("%SYSTEM%", systemLong)
-                .Replace("%SYS%", systemShort)
+                .Replace("%SYS%", systemShort.ToUpper())
```
mnadareski, line 1375: Please use `ToUpperInvariant`, it's actually a bit better

mnadareski (approved):

Issue #714, opened 2024-06-26
[Request] Add dumping tool command / flags to submission info


Deterous:
**Is your feature request related to a problem? Please describe.**
Getting info about the dumping tool flags used requires opening the redumper log. Would be nice to have this info more readily available.

**Describe the solution you'd like**
Dumping tool command/flags mentioned in the `Dumping Info:` section of the !submissionInfo.txt file

**Additional context**
Requires an update to RedumpLib. Worth timing this with the potential pseudo-tag mentioned in #713

mnadareski:
Apparently I have had `DumpingParameters` next to `DumpingProgram` for an incredibl ylong time and forgot. So it's available in RedumpLib already. It just needs to be populated in the relevant places, if possible.

Deterous:
Enabled for Redumper and DIC, the two programs that this was requested for.
Support for other dumping programs can be added where needed in the future.

PR #778, opened 2024-12-12
Add DumpingParameters for DIC and Redumper

Deterous:
An attempt to solve #714
Needs testing before merging.

### DIFF
--- a/MPF.Processors/DiscImageCreator.cs
+++ b/MPF.Processors/DiscImageCreator.cs
@@ -80,6 +80,7 @@ namespace MPF.Processors
             var dicVersion = GetCommandFilePathAndVersion(basePath, out var dicCmd);
             info.DumpingInfo!.DumpingProgram ??= string.Empty;
             info.DumpingInfo.DumpingProgram += $" {dicVersion ?? "Unknown Version"}";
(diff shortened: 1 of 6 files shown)


MPF.Processors/DiscImageCreator.cs
```diff
@@ -820,6 +821,33 @@ internal override List<OutputFile> GetOutputFiles(string? baseDirectory, string
...
+            if (string.IsNullOrEmpty(dicCmd))
+                return null;
+            if (!File.Exists(dicCmd))
+                return null;
+
+            try
+            {
+                // Dumping parameters are on the first line
+                using var sr = File.OpenText(dicCmd);
+                var line = sr.ReadLine();
```
mnadareski, line 841: Is it worth reading the whole file and not just the first line? I know it should only be one line, but still.

mnadareski (commented):
Overall looks accurate to my understanding of the two programs

MPF.Processors/DiscImageCreator.cs
Deterous: I restricted to one line just in case of some weird file with 100+ lines

mnadareski (approved):

Issue #744, opened 2024-09-03
[Request] Ignore useless volume labels


Deterous:
Currently MPF outputs ISO volume labels that are useless, e.g. `[T:VOL] Young, YOUNG (ISO)` or `[T:VOL] LIVINGFACTORY (ISO), LivingFactory (Joliet)`

TODO: Add logic to delete useless volume labels. Compare them after removing spaces/underscores and normalising case. Drop empty volume labels after this. And ensure that if only one volume label is left, don't add the brackets

e.g. the output after this should be 
`[T:VOL] Young` and ``[T:VOL] LivingFactory`

NovaSAurora:
Strongly request a toggle in options to include them anyway. I'd like to still have them and having it off by default will let the mods be happy.

mnadareski:
Is there a preferred order to the "accepted" labels? Like does Windows always take precedence over Joliet? Or if they match (ignoring case), is it always the one that includes at least one lower-case character?

Deterous:
Windows > UDF > ISO > Joliet
Honestly I've only seen bad results from the joliet label. Maybe it's coded wrong for DIC log parsing..? (my fault if so)

mnadareski:
The current formatting code doesn't take into account where the labels are coming from when formatting. It just assumes that they're organized properly. That is, there is no actual distinction between UDF, ISO, and Joliet when the formatting happens. If we go with the hierarchy, that will require special checking cases. Also worth noting that it actually doesn't do any sort of comparisons between labels at the moment.

mnadareski:
As discussed in Discord, part of the issue here is needing to rebuild the current mapping dictionary based on the "correct" labels. For a human user, this is nearly trivial to point out, regardless of what the source of the label is. For a computer, it's a bit harder, since either the key in the dictionary needs to be normalized OR a separate mapping needs to be made from the key to the "real label". Neither of these are trivial to implement and neither scales particularly well.

Deterous:
Currently, `labels` exists as a mapping of actual volume labels to zero or more filesystem names ("ISO", "Joliet", "UDF", "UNKNOWN")
It should be possible to compare the volume labels and pick the one with the better filesystem. There is no need to determine which is better through heuristics (which has case sensitivity, etc).

For example:
`labels` = a mapping of "EXAMPLE_DISC" = (ISO, Joliet) and "Example Disc" = (UDF)
1. Drop all underscores, spaces, and unicode (foreign language characters) from both labels
[...]

PR #780, opened 2024-12-13
Better deal with volume labels

Deterous:
Trying to do a minimal effort solution to #744
No testing done yet.
Also removed Joliet parsing for DIC, it seems to be broken at the moment.

### DIFF
--- a/MPF.Frontend/Tools/SubmissionGenerator.cs
+++ b/MPF.Frontend/Tools/SubmissionGenerator.cs
@@ -94,9 +94,12 @@ namespace MPF.Frontend.Tools
                 info.TracksAndWriteOffsets.ClrMameProData = null;
 
             // Add the volume label to comments, if possible or necessary
-            string? volLabels = FormatVolumeLabels(drive?.VolumeLabel, processor?.VolumeLabels);
-            if (volLabels != null)
-                info.CommonDiscInfo!.CommentsSpecialFields![SiteCode.VolumeLabel] = volLabels;
+            if (system != RedumpSystem.MicrosoftXbox && system != RedumpSystem.MicrosoftXbox360)
+            {
+                string? volLabels = FormatVolumeLabels(drive?.VolumeLabel, processor?.VolumeLabels);
+                if (volLabels != null)
+                    info.CommonDiscInfo!.CommentsSpecialFields![SiteCode.VolumeLabel] = volLabels;
+            }
 
             // Extract info based generically on MediaType
             ProcessMediaType(info, mediaType, options.AddPlaceholders);
@@ -397,12 +400,40 @@ namespace MPF.Frontend.Tools
         }
 
         /// <summary>
+        /// Simplifies a volume label into uppercase alphanumeric only string
+        /// </summary>
+        /// <param name="labels">Volume label to simplify</param>
+        /// <returns>Simplified volume label</returns>
+        private static string? SimplifyVolumeLabel(string? label)
+        {
+            if (label == null || label.Length == 0)
+                return label;
+            
+            var labelBuilder = new StringBuilder();
+            foreach (char c in label)
+            {
+                if ((c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') || (c >= '0' && c <= '9'))
+                    labelBuilder.Append(c);
+            }
+            string? simpleLabel = labelBuilder.ToString().ToUpper();
+
+            if (simpleLabel == null || simpleLabel.Length == 0)
+                return null;
+            else
+                return simpleLabel;
+        }
+
+        /// <summary>
         /// Formats a list of volume labels and their corresponding filesystems
         /// </summary>
         /// <param name="labels">Dictionary of volume labels and their filesystems</param>
         /// <returns>Formatted string of volume labels and their filesystems</returns>
         private static string? FormatVolumeLabels(string? driveLabel, Dictionary<string, List<string>>? labels)
         {
+            // Treat empty label as null
+            if (driveLabel.Length == 0)
+                driveLabel = null;
+            
             // Must have at least one label to format
             if (driveLabel == null && (labels == null || labels.Count == 0))
                 return null;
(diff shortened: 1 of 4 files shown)


MPF.Frontend/Tools/SubmissionGenerator.cs
```diff
@@ -94,9 +95,12 @@ internal static class SubmissionGenerator
                 info.TracksAndWriteOffsets.ClrMameProData = null;
 
             // Add the volume label to comments, if possible or necessary
-            string? volLabels = FormatVolumeLabels(drive?.VolumeLabel, processor?.VolumeLabels);
-            if (volLabels != null)
-                info.CommonDiscInfo!.CommentsSpecialFields![SiteCode.VolumeLabel] = volLabels;
+            if (system != RedumpSystem.MicrosoftXbox && system != RedumpSystem.MicrosoftXbox360)
```
mnadareski, line 98: There are a few unique volume labels for XGD1/2/3 discs. `GetRedumpSystemFromVolumeLabel` inside of `FormatVolumeLabels` should catch the others.

MPF.Frontend/Tools/SubmissionGenerator.cs
```diff
@@ -396,13 +400,41 @@ private static SubmissionInfo CreateDefaultSubmissionInfo(BaseProcessor processo
...
         }
 
+        /// <summary>
+        /// Simplifies a volume label into uppercase alphanumeric only string
+        /// </summary>
+        /// <param name="labels">Volume label to simplify</param>
+        /// <returns>Simplified volume label</returns>
+        private static string? SimplifyVolumeLabel(string? label)
+        {
+            if (label == null || label.Length == 0)
```
mnadareski, line 410: Since you're doing this elsewhere, why not just return `null` for this case?

MPF.Frontend/Tools/SubmissionGenerator.cs
```diff
@@ -396,13 +400,41 @@ private static SubmissionInfo CreateDefaultSubmissionInfo(BaseProcessor processo
...
+                return label;
+            
+            var labelBuilder = new StringBuilder();
+            foreach (char c in label)
+            {
+                if ((c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') || (c >= '0' && c <= '9'))
+                    labelBuilder.Append(c);
```
mnadareski, line 417: I think there's an efficient `ToUpperInvariant` that can be used on characters here. It might be easier.

MPF.Frontend/Tools/SubmissionGenerator.cs
```diff
@@ -396,13 +400,41 @@ private static SubmissionInfo CreateDefaultSubmissionInfo(BaseProcessor processo
...
+            }
+            string? simpleLabel = labelBuilder.ToString().ToUpper();
```
mnadareski, line 419: If the above comment doesn't apply, please use `ToUpperInvariant`

MPF.Frontend/Tools/SubmissionGenerator.cs
```diff
@@ -417,7 +449,22 @@ private static SubmissionInfo CreateDefaultSubmissionInfo(BaseProcessor processo
                 return driveLabel;
             }
 
-            // If only one label, don't mention fs
+            // Remove duplicate/useless volume labels
+            string? defaultLabel = SimplifyVolumeLabel(driveLabel);
+            if (defaultLabel != null && defaultLabel.Length != 0)
```
mnadareski, line 454: So you're only going to remove duplicates if the drive label could be simplified? So if there's no drive label (for some reason) but there are other labels, then those are okay to append even if they're the same?

MPF.Frontend/Tools/SubmissionGenerator.cs
```diff
@@ -417,7 +449,22 @@ private static SubmissionInfo CreateDefaultSubmissionInfo(BaseProcessor processo
...
+            {
+                List<string> keysToRemove = new List<string>();
```
mnadareski, line 456: Would you consider using something like a `Dictionary<string, List<string>>` and create a reverse map of normalized labels to their corresponding filesystems? That may allow you to do the precedence matching that you proposed before. This would also allow for key removal as you go.

MPF.Frontend/Tools/SubmissionGenerator.cs
Deterous: All Xbox360 volume labels are known and defined exactly by the PFI (and SS wave)

MPF.Frontend/Tools/SubmissionGenerator.cs
Deterous: I've chosen to specifically do this because I want ASCII only. Microsoft's functions are happy to pass Unicode variants

mnadareski: There should also be a `char.ToUpper` that can be used.

MPF.Frontend/Tools/SubmissionGenerator.cs
mnadareski: It'd be nice to get those into `GetRedumpSystemFromVolumeLabel` then :D

Deterous: You already commented some out :)

mnadareski: Oh no, `CD_ROM` and `DVD_ROM`???

mnadareski: Fine, dropping this one for now. You win this time `CD_ROM`.

MPF.Frontend/Tools/SubmissionGenerator.cs
Deterous: I've made it such that if drive label doesn't exist (and Linq is available) then it will try find the UDF label to act as the default.

MPF.Frontend/Tools/SubmissionGenerator.cs
Deterous: Something like that may be best to make this fully featured, but I'm just looking to make this sufficient for now.

mnadareski: Please leave a quick TODO in there for this future update and I'll call this gribble set.

MPF.Frontend/Tools/SubmissionGenerator.cs
mnadareski: We'll see how this ends up working :+1:

mnadareski (approved):

PR #784, opened 2024-12-18
Verify SecuROM sector count

### LINKED ISSUE #752: [Request] Check for incorrect SecuROM sector count
There is a known issue with Plextor firmwares sometimes not correctly detecting all SecuROM sectors. As a result, some submissions of PC discs to redump are missing information (usually the first sector at 00:01:74).

Each SecuROM version has a known number of sectors that should be present. MPF should parse the number of SecuROM sectors actually detected and compare it against these known numbers:

* Pre-v3 - 216
* v3 - 90
* v4 - 99
* post-v4 - 11

Where the number of sectors is any number other than the above, an error message should be popped up/added to the !submissionInfo.txt file noting a possible error and directing the dumper to the redump forums.

It is possible that this behaviour could be requested to be implemented at the dumping tool (DIC/redumper), but as this is not an issue which can be resolved by fixing either tool, and as both are currently accepted for redump submission, it would be preferable for MPF to make this check.

**Additional context**
Two forum posts describe this issue in more detail http://forum.redump.org/post/105561/#p105561 and http://forum.redump.org/topic/56767/ibm-pc-temporarily-discarded-dumps-from-duyc37/

Deterous:
Adds warning to comments if SecuROM sector count is not expected.
Fixes #752

### DIFF
--- a/MPF.Processors/Redumper.cs
+++ b/MPF.Processors/Redumper.cs
@@ -151,10 +151,12 @@ namespace MPF.Processors
                 case RedumpSystem.IBMPCcompatible:
                 case RedumpSystem.RainbowDisc:
                 case RedumpSystem.SonyElectronicBook:
-                    info.CopyProtection!.SecuROMData = GetSecuROMData($"{basePath}.log") ?? string.Empty;
+                    info.CopyProtection!.SecuROMData = GetSecuROMData($"{basePath}.log", out SecuROMScheme? secuROMScheme) ?? string.Empty;
+                    if (secuROMScheme == SecuROMScheme.Unknown)
+                        info.CommonDiscInfo!.Comments += "Warning: Incorrect SecuROM sector count" + Environment.NewLine;
 
                     // Needed for some odd copy protections
-                    info.CopyProtection!.Protection = GetDVDProtection($"{basePath}.log", false) ?? string.Empty;
+                    info.CopyProtection!.Protection += GetDVDProtection($"{basePath}.log", false) ?? string.Empty;
                     break;
 
                 case RedumpSystem.DVDAudio:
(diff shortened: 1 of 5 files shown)


MPF.Processors/Redumper.cs
```diff
@@ -673,6 +675,22 @@ private static bool IsManufacturerEmpty(string inputFilename)
...
         }
 
+        #endregion
+        
+        #region Private Enums
+
+        /// <summary>
+        /// Private enum for SecuROM scheme type
+        /// </summary>
+        internal enum SecuROMScheme
```
mnadareski, line 685: This will fly for now, but it really should live in an `Enums.cs` file in the Processors project later.

MPF.Processors/Redumper.cs
```diff
@@ -151,10 +151,12 @@ public override void GenerateSubmissionInfo(SubmissionInfo info, string basePath
                 case RedumpSystem.IBMPCcompatible:
                 case RedumpSystem.RainbowDisc:
                 case RedumpSystem.SonyElectronicBook:
-                    info.CopyProtection!.SecuROMData = GetSecuROMData($"{basePath}.log") ?? string.Empty;
+                    info.CopyProtection!.SecuROMData = GetSecuROMData($"{basePath}.log", out SecuROMScheme secuROMScheme) ?? string.Empty;
+                    if (secuROMScheme == SecuROMScheme.Unknown)
+                        info.CommonDiscInfo!.Comments += "Warning: Incorrect SecuROM sector count" + Environment.NewLine;
```
mnadareski, line 156: Looks like I was wrong, you can just use `=` in both of these places and not `+=`. I had to double check in the submission tool.

MPF.Processors/Redumper.cs
```diff
@@ -1664,9 +1682,12 @@ internal static bool GetSaturnBuildInfo(string? segaHeader, out string? buildDat
         /// Get the SecuROM data from the input file, if possible
         /// </summary>
         /// <param name="log">Log file location</param>
+        /// <param name="secuROMScheme">SecuROM scheme found</param>
         /// <returns>SecuROM data, if possible</returns>
-        internal static string? GetSecuROMData(string log)
+        internal static string? GetSecuROMData(string log, out SecuROMScheme secuROMScheme)
```
mnadareski, line 1687: Is this also something that applies to DIC?

mnadareski (approved):

Issue #832, opened 2025-03-30
[Problem] CD command renamed to DISC command in latest version of Redumper


PurpleNekoNova:
Manually changing cd to disc works in the parameters.

Issue #831, opened 2025-03-22
[Problem]: BD dumping in MPF not possible with newer version of redumper


Ragowit:
**Version**
- [x] Stable release (3.3.0)

**Build**
- [x] .NET 9.0 running on (Win 11 24H2)

**Describe the issue**
I manually updated redumper to the latest version (Build 518) and tried to dump a PS3 disc. It doesn't work, seems like redumper have removed the "bd" parameter. So I manually removed "bd" from the parameters in MPF and now it dumps, but once completed MPF says "No dumping command was run, submission information will not be gathered.", but all the files needed are dumped correctly and in the specified folder.

**To Reproduce**
1. Go to https://github.com/superg/redumper/releases
2. Download latest version (or 518 used when creating this issue)
3. Replace "redumper.exe" in "\Programs\Redumper\" with the downloaded one above.
4. Dump PS3 disc (or probably any BD disc)

**Expected behavior**
To start dump process without the need of changing parameters, and that submission information will be gathered once done.

Ragowit:
Tracked down the change in redumper to build 514, so build 508 is the latest one can manually update to in MPF.

Deterous:
Redumper has a lot of new features that MPF doesn't support yet. Stick with the bundled version of redumper, or use command line if there's something you need from mthe new Redumper

Broken-Scholar:
Build 518 completely fails to launch in the context of MPF - can confirm build 508 functions as expected in the latest stable MPF.

Deterous:
Again, highly advised not to update redumper manually if you are using MPF. The inputs/outputs have changed.
If you want to use any newer features, run redumper directly.

Issue #824, opened 2025-02-24
[Problem] Audio CD: Perfect Audio Offset of +0 incorrectly reported for all instances of Perfect Audio Offset


mictlantecuhtle:
**Version**
What version are you using?

- [X] Stable release ([version here](http://redump.org/disc/69435/))

**Build**
What runtime version are you using?

- [X] .NET 9.0 

**Describe the issue**
Wherever the phrase "Perfect Audio Offset applied" appears in redumper logs, MPF incorrectly adds a line in !submissionInfo.txt to say `<b>Ring Perfect Audio Offset</b>: +0`

This line should only be added when BOTH "Perfect Audio Offset applied" AND "disc write offset: +0" appear in the logs.

From examination of various sets of dump logs, this behaviour occurs with both positive and negative offsets.

**To Reproduce**

1. Dump an Audio CD with either a positive or negative offset.
2. Check the redumper log to verify that e.g. ""Perfect Audio Offset applied" and "disc write offset: -12" appear
3. Check the !submissionInfo.txt file to verify that the line `<b>Ring Perfect Audio Offset</b>: +0` appears

**Expected behavior**
The line `<b>Ring Perfect Audio Offset</b>: +0` should only be added when BOTH "Perfect Audio Offset applied" AND "disc write offset: +0" appear in the logs.

[122041_Foil1332 (+37 offset).zip](https://github.com/user-attachments/files/18940149/122041_Foil1332.%2B37.offset.zip)

[121855_Maddog (-12 offset).zip](https://github.com/user-attachments/files/18940171/121855_Maddog.-12.offset.zip)

Issue #823, opened 2025-02-19
[Problem] Latest Redumper version (build_481) not detected properly


Daja177:
**Version**
What version are you using?

- [x] Stable release (3.3.0-679b40de7c089c3705ef008fbb720f6a4eabc6f1)

**Build**
What runtime version are you using?

- [x] .NET 8.0.8 running on Windows 11

**Describe the issue**
When using MPF (MPF.UI) with a newer Redumper build (in this case build_481), MPF doesn't detect the Redumper version correctly.

**To Reproduce**

1. Download MPF.UI 3.3.0
2. Download Redumper build build_481 or newer
3. Replace `MPF.UI_3.3.0_net9.0-windows_win-x64_release\Programs\Redumper\redumper.exe` with `redumper.exe` gotten from build_481
4. Dump any disc using Redumper
5. Check the bottom of the generated `!submissionInfo.txt` file to find `Dumping Program: Redumper Unknown Version`

**Expected behavior**
The line should say `Dumping Program: Redumper v2025.02.17 build_481`.

**Screenshots**
![Image](https://github.com/user-attachments/assets/20822c74-8fb3-4d54-a0d6-0c393f02fd56)

**Logs**
[MPF generated logs ZIP-file](https://github.com/user-attachments/files/18873346/Sonic.Mania.Plus.Original.Soundtrack.Disc.2._logs.zip)

PR #835, opened 2025-04-17
Update redumper to build 549

### LINKED ISSUE #812: [Problem] v3.3.0 misinterpreting RedumperLeadinRetryCount setting as drive speed
**Version**

- Stable release v3.3.0 (2025-01-03)

**Build**
What runtime version are you using?

- .NET 9.0 running on Windows 11 x64

**Describe the issue**
If I enable "Set Default Plextor Lead-in Retries" under Tools > Options > Programs > Redumper, the retry count value (which defaults to 4) is passed to redumper in the speed parameter, regardless of what is chosen for the Drive Speed. If I put a custom retry count, the same number shows up in the Parameters box but is attached to the `--speed` flag, not the `--plextor-leadin-retries` argument (which is missing). Indeed, if I start the dump, redumper is invoked with `--speed=4` or similar.

**To Reproduce**

1. Start MPF.UI with a fresh config (delete config.json).
2. Set "Media Type" to `Audio CD`, observe default "Drive Speed" for CD-ROM is set to 24 and `--speed=24` appears in the "Parameters" box.
3. Change "Drive Speed" to 8, observe `--speed=8` in the "Parameters" box.
4. Enable "Set Default Plextor Lead-in Retries", observe that it now has `--speed=4`.
5. Change "Drive Speed" to 16, observe that it still shows `--speed=4`.
6. Change "Plextor Lead-in Retries" to 123, observe `--speed=123` now.

**Expected behavior**
For redumper, the `--speed` flag should be correctly set to the chosen drive speed, and the Plextor lead-in retry count value should be passed to the `--plextor-leadin-retries` flag.

Deterous:
- Use redumper build 549 for MPF publishing
- Support new commands/flags, drop removed commands/flags
- Fix minor issues with redumper processor

Fixes #812 #813 #818 #822 #823 #824 #831 #832


MPF.ExecutionContexts/Redumper/ExecutionContext.cs
```diff
@@ -225,13 +243,8 @@ public override string GenerateParameters()
...
         {
-            return ModeValues?.Contains(CommandStrings.CD) == true
-                || ModeValues?.Contains(CommandStrings.DVD) == true
-                || ModeValues?.Contains(CommandStrings.BluRay) == true
-                || ModeValues?.Contains(CommandStrings.SACD) == true
-                || ModeValues?.Contains(CommandStrings.New) == true
-                || ModeValues?.Contains(CommandStrings.Dump) == true
-                || ModeValues?.Contains(CommandStrings.DumpNew) == true;
+            return ModeValues?.Contains(CommandStrings.Disc) == true
+                || ModeValues?.Contains(CommandStrings.Dump) == true;
```
mnadareski, line 247: Does `dump::extra` need to be included here as well?

MPF.Frontend/Tools/SubmissionGenerator.cs
```diff
@@ -199,14 +199,34 @@ public async static Task<bool> FillFromRedump(Options options,
                 }
 
                 // If the line ends in a known extra track names, skip them for checking
-                if (hashData.Contains("(Track 0).bin")
+                // TODO: Smarter method to ignore all tracks that start with 0. 00. A. or AA.
```
mnadareski, line 202: Future PR? Or something that can be done with regex in this one?

publish-nix.sh
```diff
@@ -227,22 +227,22 @@ if [ $NO_ARCHIVE = false ]; then
                     cp -rfp Release_ANSI/* MPF.UI/bin/Debug/${FRAMEWORK}/${RUNTIME}/publish/Programs/Creator/
 
                     mkdir -p MPF.UI/bin/Debug/${FRAMEWORK}/${RUNTIME}/publish/Programs/Redumper
-                    cp -rfp redumper-2024.11.03_build438-win64/bin/redumper.exe MPF.UI/bin/Debug/${FRAMEWORK}/${RUNTIME}/publish/Programs/Redumper/
+                    cp -rfp redumper-2025.04.15_build549-Windows64/redumper.exe MPF.UI/bin/Debug/${FRAMEWORK}/${RUNTIME}/publish/Programs/Redumper/
```
mnadareski, line 230: Is it no longer in a `bin` subfolder? Or was it an accidental oversight to remove that part of the path?

MPF.ExecutionContexts/Redumper/ExecutionContext.cs
Deterous: dump::extra is leadin/leadout special commands only, I wouldn't consider it a proper dump command for this purpose

mnadareski: Makes sense to me.

MPF.Frontend/Tools/SubmissionGenerator.cs
Deterous: Can be done with smarter in the future, I don't want to attempt it. The current list is 99.999% of discs..

publish-nix.sh
Deterous: likely accidental

MPF.Frontend/Tools/SubmissionGenerator.cs
mnadareski: Okay, just wanted to confirm the intent of the TODO. Leaving it that way is fine by me.

PR #837, opened 2025-04-17
Change redumper default commands

Deterous:
Redumper now only allows a single mode command. Default no base command is used for redumper dumps. The implication is of removing skeleton from CD dumps. Also, drive speed is no longer a required parameter.

By setting drive speed to "0", MPF will now not add a speed parameter for redumper dumps. This results in a max speed dump.
This matches DIC where "0" means max speed.


MPF.Frontend/InterfaceConstants.cs
```diff
@@ -36,7 +36,7 @@ public static class InterfaceConstants
         /// <summary>
         /// Set of accepted speeds for all other media
         /// </summary>
-        public static List<int> Unknown => _speedValues.FindAll(s => s <= 1);
+        public static List<int> Unknown => [1]
```
mnadareski, line 39: Missing semicolon, as per the check failures.

MPF.ExecutionContexts/Redumper/ExecutionContext.cs
```diff
@@ -268,10 +267,10 @@ protected override void SetDefaultParameters(string? drivePath,
             switch (MediaType)
             {
                 case SabreTools.RedumpLib.Data.MediaType.CDROM:
-                    ModeValues = RedumpSystem switch
+                    ModeValue = RedumpSystem switch
```
mnadareski, line 270: Please keep these as `disc` for now. I understand why you want to change this, but I do not want to make that leap in the code yet.

MPF.ExecutionContexts/Redumper/ExecutionContext.cs
```diff
@@ -393,13 +395,16 @@ protected override bool ValidateAndSetParameters(string? parameters)
...
+                    case CommandStrings.FlashMT1339:
                     case CommandStrings.Subchannel:
                     case CommandStrings.Debug:
                     case CommandStrings.FixMSF:
                     case CommandStrings.DebugFlip:
                     case CommandStrings.DriveTest:
-                        ModeValues.Add(part);
+                        // Only allow one mode per command
+                        if (ModeValue != null)
+                            return false;
```
mnadareski, line 406: I think it's reasonable to just discard any additional mode values provided. An alternative is overwriting.

MPF.ExecutionContexts/Redumper/ExecutionContext.cs
```diff
@@ -393,13 +395,16 @@ protected override bool ValidateAndSetParameters(string? parameters)
...
+                        ModeValue = part;
```
mnadareski, line 407: Nitpick: Add extra newline above this one.

MPF.ExecutionContexts/Redumper/ExecutionContext.cs
mnadareski: An addendum specifically for the `CDROM` case: There no longer needs to be a switch since both "normal" CD and SACD will have the same value being assigned.

mnadareski (approved):

Ragowit:
I downloaded the latest WIP, I can't set the drive speed to 0 in the Options.

PR #842, opened 2025-05-04
Update redumper to build 565

Deterous:
- Bump redumper to build 565
- Add support for new --skeleton flag
- Use --skeleton by default for CD media type


MPF.ExecutionContexts/Redumper/SettingConstants.cs
```diff
@@ -8,6 +8,9 @@ public static class SettingConstants
         public const string EnableLeadinRetry = "RedumperEnableLeadinRetry";
         public const bool EnableLeadinRetryDefault = false;
 
+        public const string EnableSkeleton = "RedumperEnableSkeleton";
```
mnadareski, line 11: I don't see any UI updates to support this.

MPF.ExecutionContexts/Redumper/ExecutionContext.cs
```diff
@@ -295,6 +297,26 @@ protected override void SetDefaultParameters(string? drivePath,
                 this[FlagStrings.Verbose] = true;
                 (_inputs[FlagStrings.Verbose] as FlagInput)?.SetValue(true);
             }
+            if (GetBooleanSetting(options, SettingConstants.EnableSkeleton, SettingConstants.EnableSkeletonDefault))
```
mnadareski, line 300: Going with the other comment: Did you intend on making this user-toggleable?

MPF.ExecutionContexts/Redumper/SettingConstants.cs
Deterous: I wanted it to be an option in case someone didn't want skeletons for some reason, but also being a hidden option is probably better.

mnadareski: I'll fully admit I don't know why this is a hidden one, but at least it can be made visible easily in the future.

mnadareski (approved):

MPF.ExecutionContexts/Redumper/SettingConstants.cs
Deterous: Too many options makes for bad UI. I figured a hidden option would be a good compromise, still allowing it as an option for that one person who might want to disable it for all dumps (users can remove it per-dump via the command line in MPF)

PR #846, opened 2025-05-16
Add newlines to pkg info printing

Deterous:
Fixes two pkgs printing on the same line
http://forum.redump.org/post/126658/


MPF.Frontend/Tools/PhysicalTool.cs
```diff
@@ -549,7 +549,7 @@ public static bool GetBusEncryptionEnabled(Drive? drive)
                     var appPkgHeader = appPkgHeaderDeserializer.Deserialize(fileStream);
 
                     if (appPkgHeader != null)
-                        pkgInfo += $"{appPkgHeader.ContentID}";
+                        pkgInfo += $"{appPkgHeader.ContentID}" + Environment.NewLine;
```
mnadareski, line 552: We should move over to StringBuilder instead of this at some point.

mnadareski (approved):

PR #847, opened 2025-05-18
Better audio-only check

### LINKED ISSUE #810: [Request] Better Redumper Audio Check
In Redumper Processor, check the CUE for only Audio tracks rather than using `system.IsAudio()`

Deterous:
Fixes #810

### DIFF
--- a/MPF.Processors/BaseProcessor.cs
+++ b/MPF.Processors/BaseProcessor.cs
@@ -650,8 +650,7 @@ namespace MPF.Processors
                 return false;
 
             bool tracksExist = false;
-            string[] endOfLine = ["\r\n", "\n", "\r"];
-            foreach (string cueLine in cue!.Split(endOfLine))
+            foreach (string cueLine in cue!.Split(new[] { "\r\n", "\n", "\r" }))
             {
                 string line = cueLine.Trim();
                 if (line.Length == 0)
--- a/MPF.Processors/BaseProcessor.cs
+++ b/MPF.Processors/BaseProcessor.cs
@@ -657,7 +657,8 @@ namespace MPF.Processors
                 if (line.Length == 0)
                     continue;
 
-                string[] tokens = line.Split([' ', '\t'], StringSplitOptions.RemoveEmptyEntries);
+                string[] separators = [' ', '\t'];
+                string[] tokens = line.Split(separators, StringSplitOptions.RemoveEmptyEntries);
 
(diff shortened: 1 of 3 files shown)


MPF.Processors/Redumper.cs
```diff
@@ -75,20 +75,18 @@ public override void GenerateSubmissionInfo(SubmissionInfo info, string basePath
...
+                    if (IsAudio(info.TracksAndWriteOffsets.Cuesheet))
                     {
                         string universalHash = GetUniversalHash($"{basePath}.log") ?? string.Empty;
-                        info.CommonDiscInfo.CommentsSpecialFields![SiteCode.UniversalHash] = universalHash;
-                    }
-
-                    // Attempt to get the non-zero data start, if it's an audio disc
-                    if (System.IsAudio())
-                    {
+                        if (!string.IsNullOrEmpty(universalHash))
```
mnadareski, line 82: If you're adding this check, you can change the line above to be `string?` and remove the `?? string.Empty;`. Same with all of the below checks.

MPF.Processors/BaseProcessor.cs
```diff
@@ -640,6 +640,36 @@ internal static string SplitString(string? str, int count, bool trim = false)
...
+        /// Returns true if a Cuesheet consists of only Audio tracks, false otherwise 
+        /// </summary>
+        internal static bool IsAudio(string? cue)
+        {
+            // Ignore invalid inputs
+            if (string.IsNullOrEmpty(cue))
+                return false;
+
+            bool tracksExist = false;
+            foreach (string cueLine in cue!.Split(new[] { "\r\n", "\n", "\r" }, StringSplitOptions.None))
```
mnadareski, line 653: I think that `new[] { "\r\n", "\n", "\r" }` can be replaced by `["\r\n", "\n", "\r"]`

mnadareski: `StringSplitOptions.None` shouldn't be needed either, iirc.

Deterous: > I think that `new[] { "\r\n", "\n", "\r" }` can be replaced by `["\r\n", "\n", "\r"]`

That didn't compile, so I defined it as a new variable if you prefer.

Deterous: Nope nothing compiles that way so i've revert it back to `new[] { "\r\n", "\n", "\r" }`

mnadareski (approved):

PR #865, opened 2025-06-23
Better handling of Xbox/360, redumper build 613

Deterous:
- Update redumper to build 613, new --kreon-partial-ss option
- Don't print SS hash if SS is bad (has any zeroed challenge response entries)
- Combine processing of Xbox and Xbox360 (further move towards being system agnostic)
- Don't get volume label for xbox/360


MPF.Processors/Redumper.cs
```diff
@@ -219,75 +223,47 @@ public override void GenerateSubmissionInfo(SubmissionInfo info, MediaType? medi
...
-                        RemoveHeaderAndTrim($"{basePath}.manufacturer", $"{basePath}.dmi");
-                    if (!File.Exists($"{basePath}.pfi"))
-                        RemoveHeaderAndTrim($"{basePath}.physical", $"{basePath}.pfi");
-                    if (!File.Exists($"{basePath}.ss"))
-                        ProcessingTool.CleanSS($"{basePath}.security", $"{basePath}.ss");
-
-                    string xemidString = ProcessingTool.GetXeMID($"{basePath}.dmi");
-                    var xemid = SabreTools.Serialization.Wrappers.XeMID.Create(xemidString);
-                    if (xemid != null)
+                    string xemidString = ProcessingTool.GetXeMID($"{basePath}.dmi").Trim('\0');
```
mnadareski, line 238: Extra newline needed above this line

mnadareski, line 238: Is this guaranteed to fail if XMID succeeds? Or should both strings be attempted at the start  and then do an `else if` on the following line?

Deterous: For all known Xbox DMI files, the Xbox360 XeMID section is all nulls
For all known Xbox360 DMI files, the Xbox XMID section is all nulls.
I've just let it attempt both, so that if it encounters a bad DMI, it will appear with neither or both to indicate that something is wrong.

mnadareski: Wrong line. Above `string xemidString = ProcessingTool.GetXeMID($"{basePath}.dmi").Trim('\0');`

mnadareski (approved):

PR #870, opened 2025-07-08
Add mediatype to help text

JohnVeness:
I hope I'm right that mediatype isn't optional!


mnadareski:
Media type is no longer required with Check due to internal changes.

PR #875, opened 2025-07-16
Update redumper to build 631

Deterous:
Builds with redumper 631, supports new --refine-sector-mode flag


mnadareski (commented):
Just throw the new string into the tests and you're all set

mnadareski (approved):

PR #892, opened 2025-09-22
Add filters to handle Release Control output.

HeroponRikiBestest:
Removes things that are either unnecessary for the redump protection field, or outright should not be included in it.


MPF.Frontend/Tools/ProtectionTool.cs
```diff
@@ -117,6 +117,51 @@ public static async Task<Dictionary<string, List<string>>> RunProtectionScanOnPa
                 return "(CHECK WITH PROTECTIONID)";
             else if (protections.Count == 0)
                 return "None found [OMIT FROM SUBMISSION]";
+            
+            // Flag Securom-protected Release Control helpers and DFA/CA modules
+
+            string? pathPattern = null;
```
mnadareski, line 123: Everything you add should live in `SanitizeFoundProtections`

mnadareski (approved):

PR #895, opened 2025-09-26
Remove SkipMediaTypeDetection option, cleanup options window

Deterous:
`SkipMediaTypeDetection` option is now mostly obsolete, so this PR removes it.
Unlike long ago, MPF no longer uses the old third party library (no chance of freezing the UI so there is no essential need for the option).
And as of recently, only DIC execution context cares about the media type, so I have replaced the option with a check against Internal Program. Media type detection now runs only when DIC is selected.

Additionally, a couple of minor UI fixes have been made in the Options window.

### DIFF
--- a/MPF.Frontend/ViewModels/MainViewModel.cs
+++ b/MPF.Frontend/ViewModels/MainViewModel.cs
@@ -1213,13 +1213,12 @@ namespace MPF.Frontend.ViewModels
             if (defaultMediaType == MediaType.NONE)
                 defaultMediaType = MediaType.CDROM;
 
-            // If we're skipping detection, set the default value
-            if (Options.SkipMediaTypeDetection)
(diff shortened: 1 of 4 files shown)


MPF.Frontend/ViewModels/MainViewModel.cs
```diff
@@ -1213,13 +1213,12 @@ private void CacheCurrentDiscType()
             if (defaultMediaType == MediaType.NONE)
                 defaultMediaType = MediaType.CDROM;
 
-            // If we're skipping detection, set the default value
-            if (Options.SkipMediaTypeDetection)
+            // Skip detection if not using DIC (only DIC needs to know media type)
```
mnadareski, line 1216: I'd rather have the `if` removed entirely. The type is still relevant if the user chooses to change to DIC at some point in the process. We can consider removing the block or gating it behind DIC only in the future.

PR #927, opened 2025-12-15
Update Redumper to build 676 (rename asus flags to mediatek)

Deterous:
Kept the values in (and added to) the ToRedumperDriveType enum extension though


Deterous:
Good enough for a bump

mnadareski (approved):

PR #948, opened 2026-02-23
Support new redumper changes

Deterous:
I think all that's left after this is to parse Wii/GC logs, and bump redumper version packaged with MPF
Redumper still has some WIP development going on, so good to wait a bit till its settled before bumping the build number.


mnadareski (approved):
`shram` is next

PR #950, opened 2026-02-24
Repair incomplete SS files

Deterous:
This is my initial attempt at implementing a repair mechanism into MPF for incomplete Xbox/360 SS files. It is a C# implementation of my (research-only) python script here: https://github.com/Deterous/ParseXboxMetadata/blob/main/RepairSS.py

I use the built-in Cryptography library for the AES computation, and a bunch of inscrutable logic for checking and shuffling around data. I still need to test this, but in the meantime a review of the code changes would be nice.

Note: This is *necessary* for OmniDrive SS dumps, as they are always incomplete. It is also required for any Kreon redumper dumps that are incomplete (such as when dumping with --kreon-partial-ss or just an unlucky SS read that failed for some reason).

Extra note: While this logic ideally runs in redumper not MPF, implementing the AES code in redumper's codebase is currently infeasible. MPF is the next best place.

### DIFF
--- a/MPF.Processors/ProcessingTool.cs
+++ b/MPF.Processors/ProcessingTool.cs
@@ -1364,6 +1364,8 @@ namespace MPF.Processors
                 {
                     return false;
                 }
+                
+                // Map challenge ID to challenge type
                 cids.Add(dcrt[i + 1], i);
             }
 
@@ -1374,19 +1376,21 @@ namespace MPF.Processors
             else if (xgdType == 3)
                 ccrt_offset = 0x20;
 
-            // Repair challenge table
+            // Repair challenge table (challenge entries 22 and 23 are zeroed)
             int challenge_count = 0;
             for (int i = 0; i < 21; i++)
             {
+                // Offset into SS for the response type
+                int rOffset = 0x730 + i * 9;
+
                 // Cannot rebuild SS with orphan challenge ID
-                if (!cids.TryGetValue(ss[0x730 + i * 9 + 1], out int cOffset))
+                if (!cids.TryGetValue(ss[rOffset + 1], out int cOffset))
                     return false;
 
                 // Validate challenge type with response type
-                int rOffset = 0x730 + i * 9;
                 bool angle_challenge = false;
                 bool other_challenge = false;
-                switch (ss[cOffset])
+                switch (dcrt[cOffset])
                 {
                     case 0x14:
                         if (ss[rOffset] != 3)
@@ -1404,7 +1408,7 @@ namespace MPF.Processors
 
                         challenge_count += 1;
                         // Challenge must be in expected order
-                        if (challenge_count < 5)
+                        if (challenge_count > 5)
                             return false;
 
                         break;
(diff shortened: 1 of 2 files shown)


MPF.Processors/ProcessingTool.cs
```diff
@@ -1329,37 +1312,198 @@ public static bool CleanSS(byte[] ss)
...
                 return false;
 
-            switch (xgdType)
+            // Must be 21 challenge entries
+            if (ss[0x660] != 21)
+                return false;
+
+            // Setup decryptor
+#if NET20
+            using RijndaelManaged aes = new RijndaelManaged();
```
mnadareski:
```suggestion
            using var aes = new RijndaelManaged();
```

MPF.Processors/ProcessingTool.cs
```diff
@@ -1329,37 +1312,198 @@ public static bool CleanSS(byte[] ss)
...
+            // Perform decryption
+            byte[] iv = new byte[16];
+            byte[] dcrt = new byte[252];
+            bool ct01_found = false;
+            for (int i = 0; i < 240; i+=16)
+                decryptor.TransformBlock(ss, 0x304 + i, 16, dcrt, i);
+            Array.Copy(ss, 0x304 + 240, dcrt, 240, 12);
+
+            // Rebuild challenge response table
+            var cids = new Dictionary<byte, int>();
```
mnadareski:
```suggestion
            Dictionary<byte, int> cids = [];
```

MPF.Processors/Redumper.cs
```diff
@@ -316,12 +316,21 @@ public override void GenerateSubmissionInfo(SubmissionInfo info, MediaType? medi
...
-                    if (ProcessingTool.IsValidSS($"{basePath}.ss") && !ProcessingTool.IsValidPartialSS($"{basePath}.ss"))
+                    
+                    // Only record SS hash if it is valid
+                    if (ProcessingTool.IsFixedSS($"{basePath}.ss"))
                     {
                         string? ssCrc = HashTool.GetFileHash($"{basePath}.ss", HashType.CRC32);
                         if (ssCrc is not null)
                             info.CommonDiscInfo.CommentsSpecialFields[SiteCode.SSHash] = ssCrc.ToUpperInvariant();
                     }
+                    else if (ProcessingTool.FixSS($"{basePath}.ss", $"{basePath}.ss"))
```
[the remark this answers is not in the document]
Deterous: Yes, it fixes the input .ss and overwrites

MPF.Processors/ProcessingTool.cs
```diff
@@ -1329,37 +1309,210 @@ public static bool CleanSS(byte[] ss)
...
+
+            // Setup decryptor
+#if NET20
+            using var aes = new RijndaelManaged();
+            aes.BlockSize = 128;
+#else
+            using var aes = Aes.Create();
+#endif
+            aes.Key = [0xD1, 0xE3, 0xB3, 0x3A, 0x6C, 0x1E, 0xF7, 0x70, 0x5F, 0x6D, 0xE9, 0x3B, 0xB6, 0xC0, 0xDC, 0x71];
+            aes.Mode = CipherMode.CBC;
```
mnadareski: I noticed the original code was using ECB, not CBC

PR #967, opened 2026-04-24
Add Disc Hologram ID placeholders for all Xbox discs

Zopolis4:
I don't have a dotnet build setup to test this, but this should work. 

~~I don't believe original xbox discs have this ID, based on the 30 or so redump entries I checked as well as the discs I personally own?~~ apparently some do

### DIFF
--- a/MPF.Frontend/Tools/SubmissionGenerator.cs
+++ b/MPF.Frontend/Tools/SubmissionGenerator.cs
@@ -906,12 +906,19 @@ namespace MPF.Frontend.Tools
                     info.CommonDiscInfo.EXEDateBuildDate ??= addPlaceholders ? RequiredValue : string.Empty;
                     break;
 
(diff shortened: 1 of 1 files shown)


MPF.Frontend/Tools/SubmissionGenerator.cs
```diff
@@ -906,12 +906,18 @@ private static bool ProcessSystem(SubmissionInfo info, RedumpSystem? system, Dri
                     info.CommonDiscInfo.EXEDateBuildDate ??= addPlaceholders ? RequiredValue : string.Empty;
                     break;
 
+                case RedumpSystem.MicrosoftXbox360:
```
mnadareski, line 910: This can safely extend to Xbox original as well

Zopolis4: Fixed.

mnadareski (approved):

PR #971, opened 2026-04-30
Convert type of reports from success to neutral for essential steps

shfil119:
```
SUCCESS: Successfully logged into Redump!
Finding disc matches on Redump, this might take a while... 
FAILURE: No matches found for xyz
SUCCESS: Match finding complete! No matches found
```

In current form word FAILURE "disappears" between  2 SUCCESSes, imho removing them would increase readability of log from MPF.CLI. (as long as this doesn't break anything)

Just a random idea I got during dumping, feel free to close if you have different opinion.


mnadareski (approved):
Yeah, makes sense

PR #975, opened 2026-05-06
Avalonia UI

Deterous:
This PR adds an alternate UI framework for MPF based on Avalonia, allowing for Linux and macOS UI builds.
There may be some small tweaks needed before merging during my testing, but I want to open the PR now to begin the review process as this is a big change.
Thanks to 7 (@whatev-indus) for putting in the work on this, I've just helped with guidance and then cleaned it up to use SabreTools styles and build processes. I will also be responding to any PR review comments. If you want to make changes to the branch directly and cannot, ask 7 for contributor access to the repo.

This does not replace the WPF UI, currently both are built separately. The images (icon, ringcode pics) are referenced from the WPF folder, but the Strings XAML are duplicated. The new MPF.Avalonia namespace is intended to reference the MPF.Frontend namespace as much as possible, so that it can be a drop-in replacement for the old WPF UI.

It is also 7's request to have AppImage builds for Linux and other build changes required for icons and niceties on macOS, although these changes can come in a later PR (to minimize the review burden)


MPF.Frontend/Drive.cs
```diff
@@ -60,6 +60,11 @@ public class Drive
         /// <remarks>Should only be used in UI applications</remarks>
         public char? Letter => Name?[0] ?? '\0';
 
+        /// <summary>
+        /// UI-friendly display name for the drive
+        /// </summary>
+        public string DisplayName => GetDisplayName();
```
mnadareski: This can be combined with the helper method into a cached property that utilizes the `field` keyword.

MPF.Frontend/Drive.cs
```diff
@@ -120,7 +125,10 @@ private void PopulateFromDriveInfo(DriveInfo? driveInfo)
             {
                 DriveFormat = driveInfo.DriveFormat;
                 TotalSize = driveInfo.TotalSize;
-                VolumeLabel = driveInfo.VolumeLabel;
+                if (Environment.OSVersion.Platform != PlatformID.Unix)
```
mnadareski: If this is specific to Windows, then I'd prefer the check to say that instead. If saying "non-Unix" is actually the right way of doing something like Windows and Mac only, I would appreciate a one-line comment above this if/else that says that.

MPF.Frontend/Drive.cs
```diff
@@ -238,6 +246,33 @@ public void RefreshDrive()
...
+        /// <summary>
+        /// Get a human-readable string for the Drive's name
+        /// </summary>
+        private string GetDisplayName()
+        {
+            if (string.IsNullOrEmpty(Name))
+                return string.Empty;
+
+            // Deal with non-Windows drive names
+            if (Environment.OSVersion.Platform == PlatformID.Unix)
```
mnadareski: Same question/concern here with regards to the platform. Is Unix both Mac and Linux?

MPF.Frontend/Drive.cs
```diff
@@ -238,6 +246,33 @@ public void RefreshDrive()
...
+            {
+                string volumePath = Name!.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+                string volumeName = Path.GetFileName(volumePath);
+                if (!string.IsNullOrEmpty(volumeName))
+                    return volumeName;
+            }
+
+            string displayName = Name!;
+            if (displayName.Length > 1)
+                displayName = displayName.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
```
mnadareski: I don't understand why this has to be gated by a length check

MPF.Avalonia/Program.cs
```diff
@@ -0,0 +1,16 @@
...
+    internal static class Program
+    {
+        [System.STAThread]
+        public static void Main(string[] args)
+            => BuildAvaloniaApp().StartWithClassicDesktopLifetime(args);
+
+        public static AppBuilder BuildAvaloniaApp()
+            => AppBuilder.Configure<App>()
+                .UsePlatformDetect()
+                .LogToTrace();
```
mnadareski, line 25: What does `LogToTrace` do for an Avalonia application?

MPF.Avalonia/Windows/WindowBase.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            foreach (TabControl tabControl in this.GetVisualDescendants().OfType<TabControl>())
+            {
+                Thickness margin = tabControl.Margin;
+                tabControl.Margin = new Thickness(margin.Left, margin.Top + 4, margin.Right, margin.Bottom);
+            }
+        }
+
+        protected string StringResource(string key, string fallback)
+            => global::Avalonia.Application.Current?.TryFindResource(key, out object? value) == true ? value?.ToString() ?? fallback : fallback;
```
mnadareski: What is the purpose of the `global::` specifier here? This is not a pattern that has been used in this code base before and I want to make sure I understand why it's being used.

MPF.Avalonia/Windows/OptionsWindow.axaml.cs
```diff
@@ -0,0 +1,117 @@
...
+            bool? success = await RedumpClient.ValidateCredentials(
+                OptionsViewModel.Options.Processing.Login.RedumpUsername,
+                OptionsViewModel.Options.Processing.Login.RedumpPassword);
+
+            _ = MessageBoxWindow.ShowAsync(this, "Redump", OptionsViewModel.GetRedumpLoginResult(success), 1, success == true);
+        }
+
+        private void NonRedumpModeClicked(object? sender, RoutedEventArgs e)
+        {
+            if (this.FindControl<CheckBox>("NonRedumpModeCheckBox")!.IsChecked != true)
```
mnadareski, line 168: I don't know if this is enforced by the editorconfig, but, unless there is ambiguity with the location of a method or property, you can avoid using the `this.` prefix. This goes for all instances in this PR, not just this location.

MPF.Avalonia/Windows/OptionsWindow.axaml.cs
```diff
@@ -0,0 +1,117 @@
...
+            this.FindControl<Button>("DiscImageCreatorPathButton")!.Click += BrowseForDiscImageCreatorPathClick;
+            this.FindControl<Button>("RedumperPathButton")!.Click += BrowseForRedumperPathClick;
+            this.FindControl<Button>("DefaultOutputPathButton")!.Click += BrowseForDefaultOutputPathClick;
+            this.FindControl<Button>("AcceptButton")!.Click += OnAcceptClick;
+            this.FindControl<Button>("CancelButton")!.Click += OnCancelClick;
+            this.FindControl<Button>("RedumpLoginTestButton")!.Click += OnRedumpTestClick;
+            this.FindControl<CheckBox>("NonRedumpModeCheckBox")!.Click += NonRedumpModeClicked;
+        }
+
+        private async Task<string?> BrowseForExecutableAsync(string title)
```
mnadareski, line 77: Is it possible to cache the list of file picker file types if they are static per OS?

MPF.Avalonia/Windows/OptionsWindow.axaml.cs
```diff
@@ -0,0 +1,117 @@
...
+            InitializeComponent();
+            DataContext = new OptionsViewModel(options);
+            Opened += (_, _) =>
+            {
+                OptionsViewModel.Title = Title;
+                WireEvents();
+            };
+        }
+
+        private void WireEvents()
```
mnadareski: Please try to keep this as similar to the existing WPF code as possible. With very few exceptions, the click event registrations can happen in the calling site. I'm not seeing this method being used anywhere else in this file. I believe there are other files that are using this pattern as well, so this comment applies to them too.

MPF.Avalonia/Windows/OptionsWindow.axaml.cs
```diff
@@ -0,0 +1,117 @@
...
+        public OptionsWindow()
+            : this(new Options())
+        {
+        }
+
+        public OptionsWindow(Options options)
+        {
+            InitializeComponent();
+            DataContext = new OptionsViewModel(options);
+            Opened += (_, _) =>
```
mnadareski: If you want to keep with the `WireEvents` method way of handling registration, why is the event handler for `Opened` an anonymous function? If you go this route, `WireEvents` can be rewritten to be an `Opened` event handler instead. Same as the other comment, this applies to all other places that `WireEvents` methods are being used.

MPF.Avalonia/Windows/MediaInformationWindow.axaml.cs
```diff
@@ -0,0 +1,468 @@
...
+
+        private void EnsureSpecialFields()
+        {
+            MediaInformationViewModel.SubmissionInfo.CommonDiscInfo ??= new SabreTools.RedumpLib.Data.Sections.CommonDiscInfoSection();
+            MediaInformationViewModel.SubmissionInfo.CommonDiscInfo.CommentsSpecialFields ??= [];
+            MediaInformationViewModel.SubmissionInfo.CommonDiscInfo.ContentsSpecialFields ??= [];
+            MediaInformationViewModel.SubmissionInfo.Extras ??= new SabreTools.RedumpLib.Data.Sections.ExtrasSection();
+        }
+
+        private static bool ParseBooleanComment(string? value)
```
mnadareski: I believe that I have a helper for this already in IO or Serialziation. Look for references to `YesNo`.

MPF.Avalonia/Services/DialogService.cs
```diff
@@ -0,0 +1,55 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Avalonia.Controls;
+using Avalonia.Platform.Storage;
+
+namespace MPF.Avalonia.Services
```
mnadareski, line 7: General question: I see this `Service` pattern being used, but I'm unsure if this was a personal preference or something that is recommended for Avalonia. I would like to understand why these are considered services.

MPF.Avalonia/Services/ThemeService.cs
```diff
@@ -0,0 +1,187 @@
...
+        private static void ApplyWindowsTitleBarTheme(bool darkMode)
+        {
+            if (!OperatingSystem.IsWindows())
+                return;
+
+            if (global::Avalonia.Application.Current?.ApplicationLifetime is not IClassicDesktopStyleApplicationLifetime desktop)
+                return;
+
+            foreach (Window window in desktop.Windows)
+                ApplyWindowTitleBarTheme(window, darkMode);
```
mnadareski: This is a general style comment for all instances where this may be the case: `for`, `foreach`, and `while` loops, even those with only a single line in them, should use curly braces.

MPF.Avalonia/Services/ThemeService.cs
```diff
@@ -0,0 +1,187 @@
...
+using Avalonia.Controls.ApplicationLifetimes;
+using Avalonia.Controls;
+using Avalonia.Media;
+using Avalonia.Platform;
+using Avalonia.Styling;
+using MPF.Frontend;
+
+namespace MPF.Avalonia.Services
+{
+    internal static class ThemeService
```
mnadareski: Is there a reason that the existing pattern of mapped sets of themes was not used here? The colors are already define generally and helps provide a separation of how the themes are applied.

MPF.Avalonia/Services/ThemeService.cs
```diff
@@ -0,0 +1,187 @@
...
+            {
+                return null;
+            }
+        }
+
+        private const int DwmWindowAttributeUseImmersiveDarkModeBefore20H1 = 19;
+        private const int DwmWindowAttributeUseImmersiveDarkMode = 20;
+
+        [DllImport("dwmapi.dll")]
+        private static extern int DwmSetWindowAttribute(IntPtr hwnd, int attribute, ref int attributeValue, int attributeSize);
```
mnadareski: I'm fully aware that this is only in the Windows-only path, but seeing a DLL import doesn't sit entirely well. Is there any other way of setting this?

MPF.Avalonia/Services/ThemeService.cs
```diff
@@ -0,0 +1,187 @@
...
+                ?? RunCommand("kreadconfig5", "--group General --key ColorScheme");
+
+            return IsDarkThemeName(colorScheme);
+        }
+
+        private static bool IsDarkThemeName(string? value)
+            => value?.IndexOf("dark", StringComparison.OrdinalIgnoreCase) >= 0
+                || value?.IndexOf("prefer-dark", StringComparison.OrdinalIgnoreCase) >= 0;
+
+        private static string? RunCommand(string fileName, string arguments)
```
mnadareski: Similarly to my apprehension around the `DllImport` below, I'm a bit worried about this random external execution path sitting in the themes of all places.

MPF.Avalonia/Services/WindowChromeService.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            objc_msgSend_double(windowButton, sel_registerName("setAlphaValue:"), 0);
+            objc_msgSend_bool(windowButton, sel_registerName("setHidden:"), true);
+            objc_msgSend_bool(windowButton, sel_registerName("setEnabled:"), false);
+        }
+
+        private const ulong NSWindowStyleMaskMiniaturizable = 1UL << 2;
+        private const ulong NSWindowStyleMaskResizable = 1UL << 3;
+        private const long NSWindowMiniaturizeButton = 1;
+        private const long NSWindowZoomButton = 2;
```
mnadareski: This is a general style comment for all places where this applies: I strongly prefer that constants and other class-level variables are placed at the top of the file for visibility.

MPF.Avalonia/UserControls/MultiSelectDropDown.axaml.cs
```diff
@@ -0,0 +1,105 @@
+using System.Collections;
+using System.Collections.Specialized;
+using System.Linq;
+using Avalonia;
+using Avalonia.Controls;
+using Avalonia.Interactivity;
+
+namespace MPF.Avalonia.UserControls
+{
+    public partial class MultiSelectDropDown : UserControl
```
mnadareski, line 13: It is highly unfortunate that this needed to be a custom type. It definitely explains why that was one of the sticking points in previous Avalonia attempts, though.

MPF.Avalonia/Windows/CreateIRDWindow.axaml.cs
```diff
@@ -0,0 +1,117 @@
...
+        public CreateIRDWindow(MainWindow parent)
+        {
+            InitializeWindow();
+        }
+
+        private void InitializeWindow()
+        {
+            InitializeComponent();
+            DataContext = new CreateIRDViewModel();
+            Opened += (_, _) => WireEvents();
```
mnadareski: This one is a perfect example for my other comment about `WireEvents`. Instead of the event handler being the anonymous function whose only job is to call `WireEvents`, just converting `WireEvents` to use the correct parameters and calling it directly would fit existing design patterns better.

MPF.Avalonia/App.axaml.cs
```diff
@@ -0,0 +1,30 @@
...
+            AvaloniaXamlLoader.Load(this);
+            StringResourceLoader.Load(Resources, InterfaceLanguage.English);
+        }
+
+        public override void OnFrameworkInitializationCompleted()
+        {
+            ThemeService.ApplySystemTheme(Resources, OptionsLoader.LoadFromConfig());
+
+            if (ApplicationLifetime is IClassicDesktopStyleApplicationLifetime desktop)
+                desktop.MainWindow = new MainWindow();
```
mnadareski, line 34: Is there ever a possibility that the lifetime is not the classic desktop style? Or is this a general Avalonia design pattern that I am unaware of?

MPF.Avalonia/Windows/WindowBase.cs
```diff
@@ -0,0 +1,63 @@
...
+                return;
+
+            foreach (TabControl tabControl in this.GetVisualDescendants().OfType<TabControl>())
+            {
+                Thickness margin = tabControl.Margin;
+                tabControl.Margin = new Thickness(margin.Left, margin.Top + 4, margin.Right, margin.Bottom);
+            }
+        }
+
+        protected string StringResource(string key, string fallback)
```
mnadareski, line 45: This method feels like it could be broken up into a couple lines to make it easier to parse out. The combination of the ternary statement and the null coalesce makes it difficult at a glance to see what it's actually doing.

MPF.Avalonia/Windows/MediaInformationWindow.axaml.cs
```diff
@@ -0,0 +1,468 @@
...
+        }
+
+        private void EnableTabsInInputFields()
+        {
+            foreach (string name in new[]
+            {
+                "L0MasteringRing", "L0MasteringSID", "L0Toolstamp", "L0MouldSID", "L0AdditionalMould",
+                "L1MasteringRing", "L1MasteringSID", "L1Toolstamp", "L1MouldSID", "L1AdditionalMould",
+                "L2MasteringRing", "L2MasteringSID", "L2Toolstamp",
+                "L3MasteringRing", "L3MasteringSID", "L3Toolstamp",
```
mnadareski: This feels like it should be a `static readonly` collection at the class level. I have been thrown off seeing this in the declaration of a `foreach` statement multiple times now.

MPF.Avalonia/Windows/MediaInformationWindow.axaml.cs
```diff
@@ -0,0 +1,468 @@
...
+using MPF.Frontend.ViewModels;
+using SabreTools.RedumpLib.Data;
+
+namespace MPF.Avalonia.Windows
+{
+    public partial class MediaInformationWindow : WindowBase
+    {
+        private readonly bool _showPcMacHybridAlways;
+
+        private static readonly (string Name, SiteCode Code)[] CommentFields =
```
mnadareski: Why are these arrays of tuples instead of a dictionary?

MPF.Avalonia/Windows/MainWindow.axaml.cs
```diff
@@ -0,0 +1,768 @@
...
+            if (string.IsNullOrWhiteSpace(driveName) || !driveName.StartsWith("/Volumes/", StringComparison.Ordinal))
+                return;
+
+            string parameters = MainViewModel.Parameters;
+            if (string.IsNullOrWhiteSpace(parameters))
+                return;
+
+            string escapedDrive = Regex.Escape(driveName);
+            parameters = Regex.Replace(parameters, $@"(^|\s)--drive=(?:""{escapedDrive}/?""|{escapedDrive}/?)(?=\s|$)", "$1");
+            parameters = Regex.Replace(parameters, $@"(^|\s)--drive\s+(?:""{escapedDrive}/?""|{escapedDrive}/?)(?=\s|$)", "$1");
```
mnadareski: The drive names need to be escaped for Redumper? I'm having a hard time following what these regex statements are actually doing to the resulting parameters.

MPF.Avalonia/Windows/MainWindow.axaml.cs
```diff
@@ -0,0 +1,768 @@
...
+            group.Menu = menu;
+            return group;
+        }
+
+        private static NativeMenuItem CreateNativeMenuItem(string header, Action action)
+        {
+            var item = new NativeMenuItem { Header = CleanMenuHeader(header) };
+            item.Click += (_, _) => action();
+            return item;
+        }
```
mnadareski: In a similar vein to the `WireEvents` methods, is there a reason that what is being passed in isn't just a `Func` or `Action` with the correct parameters for the event handler?

MPF.Avalonia/Windows/MainWindow.axaml.cs
```diff
@@ -0,0 +1,768 @@
...
+using MPF.Frontend.ViewModels;
+using SabreTools.RedumpLib;
+using SabreTools.RedumpLib.Data;
+using AvaloniaWindowState = Avalonia.Controls.WindowState;
+
+namespace MPF.Avalonia.Windows
+{
+    public partial class MainWindow : WindowBase
+    {
+        private const double ExpandedLogPanelExtraHeight = 40;
```
mnadareski, line 35: I'm a bit surprised that dimensions are allowed to have fractional values in Avalonia.

mnadareski (commented):
Many of the inline comments pertain to multiple locations, but there were a couple of overarching comments that I wanted to mention as well:

- There are basically no code comments. For anything that was based on existing WPF code, please copy the comment wholesale. For anything that is new functionality, add at the bare minimum a `summary` statement about what it is doing. I know that many are self-explanatory once you read the method, but those comments are used to get an understanding while coding to avoid having to look at the source for all details.
- All loops, such as `for`, `foreach`, and `while` should have curly braces, regardless of the number of lines that are contained within.
- I am seeing the use of `global::` in many places and I would like to understand more about why this pattern is being used.
- Unless there is ambiguity about what method or property is being invoked, you can avoid use of the `this.` qualifier.
- There are many places, such as large lists of constants/variables or OS-specific methods, that could do with `#region` tags. This can help break up the code into more distinct sections and allows collapsing away logical groupings.

MPF.Frontend/Drive.cs
Deterous: yes

MPF.Avalonia/Program.cs
whatever-industries: `LogToTrace()` is standard Avalonia `AppBuilder` boilerplate — it routes the framework's own internal diagnostics (layout/binding/render warnings) to `System.Diagnostics.Trace`. It doesn't log anything of ours; it just surfaces framework-level warnings in the debugger output. Safe to leave in, and easy to drop if you'd rather not have it.

MPF.Avalonia/Windows/WindowBase.cs
whatever-industries: The `global::` prefix is required because our root namespace `MPF.Avalonia` shadows the framework's `Avalonia` namespace. Without it, `Avalonia.Application` resolves to our own `MPF.Avalonia.Application` instead of the framework type; `global::Avalonia.Application` forces resolution from the global root. Every `global::Avalonia...` usage exists for this single reason — the only alternative would be renaming our namespace away from `MPF.Avalonia`.

MPF.Avalonia/App.axaml.cs
whatever-industries: In this app it's always the classic desktop lifetime, so the branch is always taken. The `is IClassicDesktopStyleApplicationLifetime` check is the standard Avalonia template pattern — Avalonia supports other lifetimes (e.g. `ISingleViewApplicationLifetime` for mobile/browser targets), and the type-check keeps the same `App` class portable across them. It's a general Avalonia design pattern rather than something specific to us.

MPF.Avalonia/Services/DialogService.cs
whatever-industries: It was a personal organizational choice, not an Avalonia recommendation. These are thin static wrappers over `owner.StorageProvider` (the file/folder pickers); grouping the cross-window helpers (dialogs, theming, window chrome) under `Services/` just keeps the window code-behind from repeating the `StorageProvider` boilerplate. There's no DI or service-locator behind it. Happy to rename them (e.g. `DialogHelper`) if the `Service` naming is misleading.

MPF.Avalonia/Services/ThemeService.cs
whatever-industries: No strong reason beyond it being the most direct port. Avalonia's theming model (`ThemeVariant`/`PlatformThemeVariant` plus resource dictionaries) differs enough from the WPF approach that a 1:1 port of the mapped-set pattern wasn't clean, so the colors ended up defined inline. It can be refactored to the mapped-set style for parity and separation if you'd prefer — happy to do that.

MPF.Avalonia/Services/ThemeService.cs
whatever-industries: `DwmSetWindowAttribute` is the Win32 call that switches the title bar to dark mode on Windows 10/11, and Avalonia exposes no managed API for the immersive dark title bar — so the P/Invoke is the only option I'm aware of. It's guarded to the Windows-only path and never runs on macOS/Linux. If you'd rather not have a raw `DllImport` in the theme code, I can move it behind a small Windows-only interop helper.

MPF.Avalonia/Services/ThemeService.cs
whatever-industries: There's no cross-platform Avalonia/.NET API for reading the desktop dark-mode preference on Linux, so this queries the standard tools — `gsettings` (GNOME) and `kreadconfig6`/`kreadconfig5` (KDE). They're read-only queries with fixed arguments (no user input is interpolated, so there's no injection surface). I agree it sits oddly in the theme code; I can relocate the detection into a dedicated platform helper if you'd prefer it out of here.

MPF.Avalonia/Windows/MainWindow.axaml.cs
whatever-industries: On macOS the optical drive mounts under `/Volumes/`, and Redumper there doesn't take the `--drive` argument the way it does on Windows. These statements strip any `--drive=<vol>` / `--drive <vol>` matching the mounted volume back out of the parameter string. `Regex.Escape(driveName)` is only there to make the volume name match literally — mounted volume names can contain spaces or regex metacharacters, so without escaping the pattern could misbehave. I can add an inline comment spelling that out.

MPF.Avalonia/Windows/MainWindow.axaml.cs
whatever-industries: They are delegates — the parameters are named delegate types rather than `Func`/`Action`, but functionally the same. The view model takes these callbacks (log output, user-message display, media-info window) so it can drive those UI concerns without depending on the View layer, keeping the MVVM separation. If the preference is to type them as `Func<>`/`Action<>` directly, I can switch the signatures.

MPF.Avalonia/Windows/MainWindow.axaml.cs
whatever-industries: Avalonia measures layout in device-independent pixels as `double` throughout (Width/Height/Margin etc. are all `double`), so fractional values are expected rather than special. The constant here is a whole number — it's just typed as `double` to match the layout API.

MPF.Avalonia/UserControls/MultiSelectDropDown.axaml.cs
whatever-industries: Agreed it's unfortunate. Avalonia has no built-in multi-select / checkbox combo box that produces a summarized selection string, so this had to be a port of the WPF custom control. It exposes `Label`/`ItemsSource`/`SummaryText` as `StyledProperty` for binding and reflectively reads an `IsChecked` property off each bound item, so it works with whatever item type the view models supply.

MPF.Avalonia/Windows/MediaInformationWindow.axaml.cs
whatever-industries: They're kept as ordered arrays because display order matters — the comment/content fields render in a fixed sequence, and a `Dictionary` doesn't guarantee enumeration order, so switching would risk shuffling the UI. They're also small, fixed lists iterated linearly, so a dictionary's lookup benefit doesn't really apply here. If keyed lookup is needed elsewhere I could add a dictionary alongside, but the ordered array matches how they're consumed.

MPF.Avalonia/Windows/OptionsWindow.axaml.cs
whatever-industries: Yes — those picker lists are static per OS, so they can be cached (hoisted to `static readonly`, or built once per OS) rather than allocated on each browse call. Happy to make that change if you'd like it in this PR.

MPF.Avalonia/Windows/OptionsWindow.axaml.cs
whatever-industries: Agreed — the anonymous `Opened` handler exists only to call `WireEvents`, so `WireEvents` could just be the `Opened` event handler directly to match the existing pattern. Happy to align this and the other `WireEvents` sites with that approach.

MPF.Avalonia/Services/WindowChromeService.cs
whatever-industries: Done (commit 536e1ee2) — moved the native-interop `const` values to the top of the class in both `WindowChromeService` and `ThemeService`. The `[DllImport]` externs stay grouped in their `Native Interop` region, but the constants now sit at the top for visibility as you prefer.

MPF.Avalonia/Windows/MediaInformationWindow.axaml.cs
whatever-industries: Done (commit 536e1ee2) — extracted to `static readonly` class-level arrays so they're no longer declared inline in the `foreach`: `TabEnabledFieldNames` here, and `LanguageMenuItemNames` for the equivalent block in `MainWindow`.

MPF.Avalonia/Windows/OptionsWindow.axaml.cs
whatever-industries: Done (commit 536e1ee2) — `WireEvents` is now registered directly as the `Opened` handler (`Opened += WireEvents;` with a `(object?, EventArgs)` signature) in `OptionsWindow`, `CheckDumpWindow`, and `CreateIRDWindow`, instead of an anonymous lambda. `MainWindow` already used a named `OnOpened` handler that calls `WireEvents` alongside its other first-load setup, so it was left as-is.

MPF.Avalonia/Windows/OptionsWindow.axaml.cs
whatever-industries: I've taken the route from your other comment (commit 536e1ee2): `WireEvents` is now wired directly as the `Opened` handler rather than through an anonymous lambda. I kept it as a single method rather than inlining the registrations at the call site because the existing windows centralize their hookup on `Opened`. If you'd still prefer the click registrations inlined at the call site to mirror the WPF structure more closely, I'm happy to make that change too — just let me know.

MPF.Avalonia/Windows/CreateIRDWindow.axaml.cs
whatever-industries: Done (commit 536e1ee2) — converted exactly as suggested: `Opened += WireEvents;` with `WireEvents(object? sender, EventArgs e)` as the handler, dropping the anonymous function.

MPF.Avalonia/Windows/OptionsWindow.axaml.cs
whatever-industries: I looked into removing these, but every `this.` usage in the PR is a call to an Avalonia extension method — `FindControl` (×135), `GetVisualRoot` (×1), and `GetVisualDescendants` (×1). Extension methods can't be invoked on an implicit `this`, so dropping the prefix produces `CS0103` and won't compile. There are no instance-member `this.` usages to remove, so unfortunately there's nothing to change here.

MPF.Avalonia/Windows/MainWindow.axaml.cs
whatever-industries: Added the inline comments spelling this out (commit 548f0ff6): one note on `Regex.Escape` (mounted volume names under `/Volumes/` can contain spaces or regex metacharacters, so the name is escaped to match literally), and one on the two `Regex.Replace` calls (they strip the mounted drive's `--drive=<name>` / `--drive <name>` argument — quoted or unquoted, optional trailing slash — because Redumper on macOS doesn't take the drive argument the way it does on Windows).

MPF.Frontend/Drive.cs
whatever-industries: You're right that this is effectively Windows-only. `DriveInfo.VolumeLabel` is only available on Windows, and modern .NET reports both Linux and macOS as `PlatformID.Unix`, so `!= PlatformID.Unix` resolves to "Windows" here. I didn't switch to `OperatingSystem.IsWindows()` because this project also targets net20-net48, where that helper doesn't exist. I've added a comment above the if/else explaining exactly that (commit 548f0ff6).

whatever-industries: Updated to make the check explicit (commit b76b4cb9): switched `!= PlatformID.Unix` to `== PlatformID.Win32NT`, which reads as "Windows" directly as you preferred. I used `PlatformID.Win32NT` rather than `OperatingSystem.IsWindows()` because that helper only exists on net5.0+, and this project also targets net20-net48; `PlatformID.Win32NT` is available on all of them. The comment is now a single line noting that `DriveInfo.VolumeLabel` is Windows-only.

MPF.Avalonia/Windows/MainWindow.axaml.cs
whatever-industries: To be clear on the two parts: the `Regex.Escape` and what the `Regex.Replace` calls do are as described above (escaping is only to match the volume name literally in the .NET regex; the replaces strip the mounted-volume `--drive` argument). The underlying reason the argument is stripped specifically on macOS — i.e. exactly how Redumper expects the drive to be specified there — is best confirmed by @Deterous, who authored this method; my explanation of that part is an inference rather than something I can state authoritatively.

MPF.Avalonia/Services/DialogService.cs
mnadareski: I'll leave this for now and then reorganize locally later. I don't really agree with these being called "Services", but that's a personal thing.

MPF.Avalonia/Services/ThemeService.cs
mnadareski: I would prefer the same mapped sets of themes reflected here. It may not be as clean as this, but it reduces the amount of magic numbers that appear in multiple places in the same file.

MPF.Avalonia/Services/ThemeService.cs
mnadareski: I'd like to see an example of partial classes that have OS-specific functionality. I think it may fit the needs here better than having them all in regions in the same file.

MPF.Avalonia/Services/ThemeService.cs
mnadareski: As mentioned in above response, having partial classes for each supported OS would make this easier to keep track of, even if we can't get rid of the OS-specific calls.

MPF.Avalonia/Services/WindowChromeService.cs
```diff
@@ -0,0 +1,99 @@
...
+using Avalonia.Controls;
+using Avalonia.Threading;
+
+namespace MPF.Avalonia.Services
+{
+    /// <summary>
+    /// Applies platform-specific window chrome adjustments, such as hiding unavailable
+    /// macOS title bar buttons and theming the title bar
+    /// </summary>
+    internal static class WindowChromeService
```
mnadareski: Recommendation: Rename this class to more clearly show that all of this code is MacOS-specific.

MPF.Avalonia/Windows/MainWindow.axaml.cs
```diff
@@ -0,0 +1,1011 @@
...
+
+        /// <summary>
+        /// Extra window height added when the log panel is expanded
+        /// </summary>
+        private const double ExpandedLogPanelExtraHeight = 40;
+
+        /// <summary>
+        /// Menu item control names for the selectable interface languages
+        /// </summary>
+        private static readonly string[] LanguageMenuItemNames =
```
mnadareski, line 40: A bit pedantically, but I would prefer that each of the items be on their own line within the declaration. This is a style request that has nothing to do with actual functionality.

MPF.Avalonia/Windows/MainWindow.axaml.cs
mnadareski: Typing them would make this a bit easier to follow, as the current model obscures what is actually going on. Possibly renaming the `action` parameter to be something more explicit would be good as well. Something like `onClick`.

MPF.Avalonia/Windows/MainWindow.axaml.cs
```diff
@@ -0,0 +1,1011 @@
...
+using SabreTools.RedumpLib;
+using SabreTools.RedumpLib.Data;
+using AvaloniaWindowState = Avalonia.Controls.WindowState;
+
+namespace MPF.Avalonia.Windows
+{
+    /// <summary>
+    /// Interaction logic for MainWindow.axaml
+    /// </summary>
+    public partial class MainWindow : WindowBase
```
mnadareski, line 28: Following some of the feedback in the services, if there's OS-specific code, let's try having those bits be in OS-specific partial classes. I noticed at least a couple of places that were MacOS-only.

MPF.Avalonia/Windows/MediaInformationWindow.axaml.cs
mnadareski: I don't think that the order of this collection really matters since only the order of the elements in the axaml file would affect anything. 

In the original WPF code, there was the ability to set a binding directly to the model instead of having to do this. If that's possible instead of these load and save loops, I'd prefer it.

MPF.Avalonia/Windows/MediaInformationWindow.axaml.cs
mnadareski: This is still relevant. The `SabreTools.IO` and `SabreTools.Serialization` namespaces have a bunch of helpers. If there isn't one that can handle this, that's fine.

MPF.Avalonia/Windows/OptionsWindow.axaml.cs
mnadareski: I'll strongly prefer a static readonly collection that's initialized this same way. It makes a lot more sense to me, especially since the set of patterns really doesn't change.

MPF.Avalonia/Windows/WindowBase.cs
mnadareski: This comment is still relevant. Either reformatting the line to be more clear about the ternary statement or splitting it into multiple lines would be preferred.

MPF.Avalonia/Services/ThemeService.cs
whatever-industries: Done (commit fdcee97c) — the inline colors are now mapped sets of themes mirroring the WPF pattern: an abstract `Theme` base class (`MPF.Avalonia/Theme.cs`) with the brush properties and an `Apply(IResourceDictionary)` mapping, plus `Themes/DarkModeTheme.cs` and `Themes/LightModeTheme.cs` defining the colors once in their constructors. `ThemeService.Apply` now just selects the theme, applies the custom background/text overrides via `Theme.ApplyCustomColors`, and applies it — no more magic numbers in the service.

MPF.Avalonia/Services/ThemeService.cs
whatever-industries: Done (commit fdcee97c) — `ThemeService` is now a partial class split by OS: `ThemeService.cs` keeps the shared theme resolution/application, `ThemeService.Linux.cs` holds the Linux dark-mode detection (the gsettings/kreadconfig queries and `RunCommand` now live there), and `ThemeService.Windows.cs` holds the DWM title bar theming.

MPF.Frontend/Drive.cs
mnadareski: This is still relevant. Combining the `GetDisplayName` method into the property itself is preferred for this small of an initializer.

MPF.Avalonia/Services/ThemeService.cs
whatever-industries: Done (commit fdcee97c) — the `DwmSetWindowAttribute` P/Invoke and the title bar theming code now live in their own `ThemeService.Windows.cs` partial, alongside a `ThemeService.Linux.cs` partial for the Linux detection, so each OS-specific path is isolated in its own file rather than regions.

MPF.Avalonia/Services/WindowChromeService.cs
whatever-industries: Done (commit fdcee97c) — renamed to `MacOSWindowChromeService`. I also moved the one non-macOS call it contained (`ThemeService.ApplyWindowTitleBarTheme`, which is Windows-only) out to the call sites in `WindowBase` and `MessageBoxWindow`, so everything in the class is now genuinely macOS-specific.

MPF.Avalonia/Windows/MainWindow.axaml.cs
whatever-industries: Done (commit fdcee97c) — `LanguageMenuItemNames` now has one item per line, and I applied the same formatting to `TabEnabledFieldNames` in `MediaInformationWindow` (and moved it up into the `Field Mappings` region at the top of the class, per your earlier comment about class-level variable placement).

MPF.Avalonia/Services/DialogService.cs
whatever-industries: Understood — leaving the naming as-is for now. (Note the title bar button hiding class formerly in here was renamed to `MacOSWindowChromeService` per your other comment, but it still lives under `Services/` pending your reorganization.)

MPF.Frontend/Drive.cs
mnadareski: Still the same question here. I think the length check can be skipped.

mnadareski:
The failing build is due to Redump login settings going from `Redump` to `RedumpOrg` naming.

MPF.Avalonia/Windows/MainWindow.axaml.cs
whatever-industries: Done (commit 20a8896e) — `CreateNativeMenuItem` now takes a properly-typed `EventHandler<RoutedEventArgs> onClick` plus the `sender` to pass through, so call sites pass the click handlers as method groups directly (`CreateNativeMenuItem(header, AppExitClick, this)`) instead of wrapping them in lambdas. The sender parameter is needed because the language items pass their corresponding `MenuItem` so `LanguageMenuItemClick` can identify which language was chosen.

MPF.Avalonia/Windows/MainWindow.axaml.cs
whatever-industries: Done (commit 20a8896e) — following the ThemeService pattern, the macOS-only parts of `MainWindow` now live in a `MainWindow.MacOS.cs` partial: the native application menu construction and header refresh (including the `_native*` menu item fields), the `/Volumes/` output path normalization, and the Redumper drive-argument stripping. `ConfigurePlatformChrome` stayed in the main file since it branches across all three platforms.

MPF.Avalonia/Windows/MediaInformationWindow.axaml.cs
whatever-industries: Done (commit a2198c12) — it is possible, and it's now implemented the same way as the WPF window. Avalonia doesn't support WPF's `[(redump:SiteCode)Key]` cast syntax in binding paths, but its binding engine converts plain string indexer arguments to the indexer's enum parameter type, so `Text="{Binding SubmissionInfo.CommonDiscInfo.CommentsSpecialFields[AlternativeTitle], Mode=TwoWay}"` works directly. All the mapped comment/content fields, `Extras.DiscKey`/`DiscID`, and the PC/Mac hybrid checkbox are now bound in the markup, and the load/save loops, the mapping arrays, and the get/set helpers are gone (~175 lines of code-behind removed). The special-field dictionaries are created before the DataContext is assigned so the two-way bindings always have a write target. I verified the full read/write round-trip against the real window under Avalonia.Headless (values flow in from the model, edits flow back, missing keys are created on write, and empty sections don't lose edits). One note for the future: this relies on reflection bindings — if the project ever migrates to compiled bindings (`x:DataType`), the enum-keyed indexer paths would need revisiting.

MPF.Avalonia/Windows/MediaInformationWindow.axaml.cs
whatever-industries: I found the helper: `ToYesNo(this string?)` in `SabreTools.RedumpLib.Data` (Extensions), which maps "yes"/"true" to `YesNo.Yes`. As it turns out, though, the direct-binding change from your other comment (commit a2198c12) removed `ParseBooleanComment` entirely — the PC/Mac hybrid checkbox is now two-way bound to the comments dictionary exactly like the WPF version, with the framework handling the string/bool conversion, so there's no manual parsing left for the helper to replace.

MPF.Avalonia/Windows/OptionsWindow.axaml.cs
whatever-industries: Done (commit 20a8896e) — the lists are now `static readonly` collections at the top of the class, initialized the same way: `WindowsExecutableFileTypes` and `AllFilesFileTypes` here, and the browse method just selects between them by OS. I applied the same treatment to the other static picker lists for consistency: `InputFileTypes` in `CheckDumpWindow`, `IrdFileTypes` in `CreateIRDWindow`, and `OutputFileTypes` in `MainWindow`. (The one in `CreateIRDWindow.OpenFileAsync` builds its filter from runtime parameters, so it stays inline.)

MPF.Avalonia/Windows/WindowBase.cs
whatever-industries: Done (commit 20a8896e) — split into multiple lines with early returns: no current application returns the fallback, a failed resource lookup returns the fallback, and only then is the resolved value converted with the null-coalesce. No more nested ternary.

whatever-industries:
Thanks for the pointer — that was exactly it. The branch was based on a master from before the `Redump` → `RedumpOrg` rename (9e6e0f62), so it built locally but failed once CI merged it with current master. I've merged master into the branch and updated the MPF.Avalonia references (commit 8004ea37): the credential test handler in `OptionsWindow.axaml.cs` now uses `RedumpOrgUsername`/`RedumpOrgPassword`/`GetRedumpOrgLoginResult`, and I also updated the two login `TextBox` binding paths in `OptionsWindow.axaml` — those wouldn't have failed the compile, but they would have silently broken the username/password fields at runtime after the rename. Full solution builds clean against current master and all 1,817 tests pass.

One parity note from the merge: master also removed the split hash fields (Size/CRC-32/MD5/SHA-1) from the WPF `MediaInformationWindow` (e39f4b5c). The Avalonia window still has its equivalents of those fields — happy to mirror that removal here if you'd like it in this PR.

mnadareski:
> One parity note from the merge: master also removed the split hash fields (Size/CRC-32/MD5/SHA-1) from the WPF MediaInformationWindow (https://github.com/SabreTools/MPF/commit/e39f4b5c9fe5c4e4c6997f7c08f197e6097ba91f). The Avalonia window still has its equivalents of those fields — happy to mirror that removal here if you'd like it in this PR.

Removing those isn't strictly necessary as a part of this PR, but it would help the cleanup effort after.

whatever-industries:
Done (commit 2183ca75) — removed the Size/CRC-32/MD5/SHA-1 fields and their collapse logic from the Avalonia `MediaInformationWindow`, mirroring e39f4b5c.

MPF.Frontend/Drive.cs
whatever-industries: Done (commit cd446612) — `GetDisplayName` is now folded into the `DisplayName` property itself as a getter cached with the `field` keyword, so the name is computed once per drive on first access. Compiles clean across all target frameworks (net20 through net10.0) since `field` is a compiler-level feature.

MPF.Frontend/Drive.cs
whatever-industries: Done (commit cd446612) — the length check is gone; the name is trimmed unconditionally now. You were right that it was redundant: the only case it protected (a single-character name like `/` trimming to empty) is already covered by the empty-string fallback that returns the raw `Name`.

mnadareski (approved):

PR #977, opened 2026-05-27
Complete macOS app for MPF: signed universal .app with auto-signed dumping tools

knutwurst:
# Complete macOS app for MPF

MPF's GUI is WPF/WinForms and therefore Windows-only. This PR delivers a **complete, ready-to-run macOS application**: a signed, universal `.app` bundle containing the full MPF UI, with native macOS window controls, light/dark themes, and the bundled dumping tools made runnable out of the box. The UI is built with [Avalonia](https://avaloniaui.net/) and **reuses the existing `MPF.Frontend` view-models unchanged** — it's a second View layer over the same MVVM core, not a logic rewrite.

> **Relationship to #975:** #975 is the primary cross-platform Avalonia-UI effort and is further along. This PR's core UI port overlaps with it, but its real focus is the **complete macOS app** — the signed `.app`, dock icon, native chrome, and on-launch signing of the bundled tools — i.e. the "icons and niceties on macOS" that #975 mentions deferring to a later PR. I'm glad to rebase just those macOS-specific pieces on top of #975 and contribute them there instead of landing a parallel UI (see the discussion below).

## Screenshots

| | Light | Dark |
|---|---|---|
| **Main window** | Main – light | Main – dark |
| **Options** | Options – light | Options – dark |

## The macOS app

- **Signed universal (arm64 + x64) `.app`** — one bundle runs natively on Apple Silicon and Intel; built by `MPF.UI.Avalonia/build-macos-app.sh`.
- **App icon** wired for the window and the macOS dock/Finder.
- **Native window decorations** (real macOS traffic-light controls, native drag/zoom/close), the OS-style **system accent**, and the **native system font** (San Francisco) so it doesn't feel like a Windows port.
- **Zero-friction dumping tools:** on launch the app strips the quarantine flag from, ad-hoc code-signs, and adds the needed `@executable_path/lib` rpath to the bundled tools under a sibling `Programs/` folder — so `redumper`/`DiscImageCreator` run on Apple Silicon with no manual Gatekeeper/`codesign` steps.
- **Live tool output in the log:** the dumping program's `stdout`/`stderr` (incl. `redumper`'s carriage-return progress) is streamed into the in-app log. Opt-in per frontend, so Windows/WPF keeps its console behavior.

## Full UI parity

- Every window is ported — Main, Options, Media Information, Check Dump, Create IRD and the Ring Code Guide — plus the `LogOutput`/`UserInput` controls and a message-box replacement.
- Binds directly to the unchanged `MPF.Frontend` view-models (`MainViewModel`, `OptionsViewModel`, `MediaInformationViewModel`, `CheckDumpViewModel`, `CreateIRDViewModel`).
- Light & dark themes; all 12 interface languages ported and switchable at runtime.

## Build & CI

- `publish-nix.sh` builds and zips `MPF.UI.Avalonia` for every desktop runtime using the **same naming scheme** as the existing projects (`MPF.UI.Avalonia_<framework>_<runtime>_<config>.zip`), bundling the dumping tools under `Programs/`.
- The Build & Test workflow gains a `workflow_dispatch` trigger and a separate **macOS-runner job** that publishes the signed universal `.app` and attaches it to the rolling release. *(That job uses macOS CI minutes — happy to gate it behind `workflow_dispatch`/tags if you prefer.)*

## macOS-specific fixes to shared code

Each also benefits the CLI on macOS:

- `DefaultDiscImageCreatorPath` now detects macOS via `RuntimeInformation` (`OSVersion.Platform` returns `Unix` on macOS, so it previously picked the Linux `.out` binary name).
- Default output names use the volume name instead of the full mount path (no more `_Volumes_DVD_ROM`).
- On macOS, `config.json` is stored under `~/.config/mpf` rather than inside the app bundle.

WPF-only constructs were replaced inside the new project only (`CustomMessageBox` → a small Avalonia dialog, WinForms file dialogs → `StorageProvider`, `System.Windows.Media` → `Avalonia.Media`, `FindResource` → Avalonia resource lookups, etc.).

## Notes for review

- The screenshots above are committed under `docs/screenshots/` purely for this description and can be moved to attachments before merge.
- The macOS-runner CI job is optional and easy to drop or gate.
- The `rpath` auto-fix uses `install_name_tool` (Xcode Command Line Tools); if those are absent, signing still works and only the rpath fix is skipped.

---

macOS app contributed by **Knutwurst** — https://github.com/knutwurst/MPF


mnadareski:
There is another PR already opened for this functionality: https://github.com/SabreTools/MPF/pull/975

knutwurst:
Thanks — and apologies for the overlap, I hadn't seen #975 when I started. You're right that #975 is the primary Avalonia-UI effort and it's further along.

Where this PR adds something is the **complete macOS app experience** — essentially the "icons and niceties on macOS" that #975 mentions deferring to a later PR:

- a signed **universal (arm64 + x64) `.app`** with a dock icon and native window controls;
- on-launch **quarantine removal + ad-hoc signing + `@executable_path/lib` rpath fix** of the bundled dumping tools, so `redumper`/`DiscImageCreator` run on Apple Silicon with no manual Gatekeeper/`codesign` steps;
- the dumping tool's live output streamed into the in-app log; and
- a few macOS-specific fixes (DiscImageCreator default path, config location, output naming from the volume label).

I'd be glad to **rebase just these macOS-specific pieces on top of #975** and contribute them there (or coordinate with @whatev-indus / @Deterous) rather than land a parallel UI — whatever means the least duplicate work for you. Happy to align with #975.

Deterous:
> Where this PR adds something is the **complete macOS app experience** — essentially the "icons and niceties on macOS" that #975 mentions deferring to a later PR:

The reason for deferring to a later PR is to minimize the review burden.

MPF.Frontend/Drive.cs
```diff
@@ -121,6 +121,17 @@ private void PopulateFromDriveInfo(DriveInfo? driveInfo)
                 DriveFormat = driveInfo.DriveFormat;
                 TotalSize = driveInfo.TotalSize;
                 VolumeLabel = driveInfo.VolumeLabel;
+
+                // On Unix/macOS DriveInfo.VolumeLabel is the mount path (e.g. "/Volumes/DVD_ROM"),
```
Deterous, line 138: Actually redumper needs the BSD drive path for dumping, not the mount path

.github/workflows/build_and_test.yml
```diff
@@ -51,9 +52,57 @@ jobs:
...
+                      [Windows x64 UI (Avalonia) Release](https://github.com/SabreTools/MPF/releases/download/rolling/MPF.UI.Avalonia_net10.0_win-x64_release.zip)"
                   name: "Rolling Release"
                   prerelease: True
                   replacesArtifacts: True
                   tag: "rolling"
                   updateOnlyUnreleased: True
+
+    # Builds the signed, universal (arm64 + x64) macOS .app on a macOS runner (needs lipo +
+    # codesign, which only exist on macOS) and appends it to the same rolling release.
+    build-macos-app:
```
Deterous, line 76: Building is preferred in the build script rather than GHA

MPF.Frontend/Drive.cs
knutwurst: Good catch. Fixed in 2e4d668.

Added a new `Drive.DevicePath` property that resolves the BSD device node via `diskutil info` on macOS, with a fallback to the mount path if diskutil isn't available. `Drive.Name` stays the mount path so the existing filesystem detection (`Directory.Exists`, `Path.Combine` on disc contents) keeps working unchanged. The execution contexts for Aaru, DiscImageCreator and Redumper now get `DevicePath` instead of `Name`, so redumper sees `/dev/disk2` rather than `/Volumes/DVD_ROM`.

.github/workflows/build_and_test.yml
knutwurst: Makes sense, moved into the script in 3fee7c3. `build-macos-app.sh` now also produces `MPF.UI.Avalonia_macos_<mode>_release.zip` next to the `.app`. Naming follows what `publish-nix.sh` uses for the other targets. The macOS job just runs the script and uploads the file.

mnadareski:
I am going to be placing my review effort on the existing PR with regards to all of the Avalonia UI. I definitely will be interested in the Mac build script and that additional setup as well as any improvements that can be made.

I request that you do not make too many changes in your existing PR as I will likely request a stripped-down PR with just the unique functionality from here.

mnadareski:
The other PR has been merged. If there are things that can be improved, I will suggest opening a separate PR based on the now-existing code.

PR #979, opened 2026-06-18
Resolve bare dumping-tool names via runtime directory and PATH

gmipf:
Resolves a configured dumping-tool path through the runtime directory and
`$PATH` when the value is a bare executable name, so distro-installed tools
(e.g. `/usr/bin/redumper`) are found without an absolute path in `config.json`.

### What this PR does
- Adds `FrontendTool.ResolveBinaryPath(string?)`: a value containing a path
  separator keeps the existing `File.Exists` behavior; a bare name (no
  separator) is looked up in the runtime directory first, then each `$PATH`
  entry.
- Wires it into the three existing call sites: the CLI (`BaseFeature`) and the
  GUI/ViewModel program-availability check and dump-environment setup
  (`MainViewModel`).
- Implements the design @mnadareski accepted in #856: "As long as it also
  looked at the runtime directory as well, then that is a reasonable
  alternative."

### What this PR doesn't do
- No assumed or hardcoded binary names — only the value the user already
  configured is resolved.
- No new environment variables (e.g. `MPF_*_PATH`).
- No PATHEXT handling: a bare name on Windows is matched verbatim. Default
  configs use path separators and hit the unchanged branch.
- No new dependencies.

### Tests
`MPF.Frontend.Test/Tools/FrontendToolTests.cs` — 8 cases: null/empty,
absolute-existing, absolute-missing, relative-with-separator missing, bare
name in runtime dir, bare name on `$PATH`, bare name not found.

Built net20 → net10.0, all green; full MPF.Frontend.Test suite passes (510).

This revives the approach from @leo60228's earlier `resolve-binary-path` PR,
extended to the CLI and ViewModel call sites. Happy to split, rebase, or strip
this down further if any of it is out of scope.

Prepared with AI assistance (Claude Opus 4.8) and reviewed before submission.

### DIFF
--- a/MPF.Frontend/Tools/FrontendTool.cs
+++ b/MPF.Frontend/Tools/FrontendTool.cs
@@ -1,4 +1,5 @@
 using System;
+using System.IO;
 using System.Reflection;
 using System.Text;
 using SabreTools.RedumpLib.Data;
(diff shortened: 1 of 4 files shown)


MPF.Frontend/Tools/FrontendTool.cs
```diff
@@ -106,6 +107,86 @@ public static int GetDefaultSpeedForMediaType(MediaType? mediaType, Options opti
...
+        /// a path with separators is treated as the explicit location it always has been; a bare
+        /// name opts the user in to a runtime-dir / <c>$PATH</c> lookup.
+        /// </remarks>
+        public static string? ResolveBinaryPath(string? configuredPath)
+        {
+            if (string.IsNullOrEmpty(configuredPath))
+                return null;
+
+            // Explicit location (absolute or relative path) — keep historical behavior.
+            if (configuredPath!.IndexOf('/') >= 0 || configuredPath.IndexOf('\\') >= 0)
```
mnadareski: I think you can use `.Contains` here instead of the `IndexOf` checks.

MPF.Frontend/Tools/FrontendTool.cs
```diff
@@ -106,6 +107,86 @@ public static int GetDefaultSpeedForMediaType(MediaType? mediaType, Options opti
...
+        /// </list>
+        /// </para>
+        /// </returns>
+        /// <remarks>
+        /// Matches the design agreed on in
+        /// <see href="https://github.com/SabreTools/MPF/pull/856">#856</see>:
+        /// a path with separators is treated as the explicit location it always has been; a bare
+        /// name opts the user in to a runtime-dir / <c>$PATH</c> lookup.
+        /// </remarks>
+        public static string? ResolveBinaryPath(string? configuredPath)
```
mnadareski, line 123: These remarks can be removed

MPF.Frontend/Tools/FrontendTool.cs
```diff
@@ -106,6 +107,86 @@ public static int GetDefaultSpeedForMediaType(MediaType? mediaType, Options opti
 
         #endregion
 
+        #region Binary Resolution
+
+        /// <summary>
+        /// Resolve a configured dumping-tool path to an absolute file path.
+        /// </summary>
+        /// <param name="configuredPath">Raw value from the user's options (Aaru/DIC/Redumper path)</param>
```
mnadareski: I think it's fine to remove the mentions of Aaru, DIC, and Redumper

MPF.Frontend/Tools/FrontendTool.cs
```diff
@@ -106,6 +107,86 @@ public static int GetDefaultSpeedForMediaType(MediaType? mediaType, Options opti
...
+        /// <returns>
+        /// The absolute path of the located binary, or <c>null</c> when nothing exists.
```
mnadareski: I don't tend to use code blocks for printing null, so those can be removed.

MPF.Frontend/Tools/FrontendTool.cs
```diff
@@ -106,6 +107,86 @@ public static int GetDefaultSpeedForMediaType(MediaType? mediaType, Options opti
...
+
+            return null;
+        }
+
+        /// <summary>
+        /// Directory containing the running application; used as the first lookup root for
+        /// bare-name binaries (see <see cref="ResolveBinaryPath(string?)"/>). Matches the
+        /// multi-TFM pattern already used in <c>OptionsLoader.GetRuntimeConfigurationPath</c>.
+        /// </summary>
+        private static string GetRuntimeDirectory()
```
mnadareski: Just check quickly that this isn't already a helper in SabreTools.IO somewhere. Maybe like `PathTool`?

gmipf:
Thanks — all addressed in aa6acb4e:

- `PathTool.GetRuntimeDirectory()` is a drop-in for the local helper, so I switched to it and removed the helper (and its multi-TFM `#if` block).
- Using `string.Contains` for the separator check.
- Trimmed the XML doc (dropped the `remarks` block, the tool-name mention, and the code formatting around `null`).

mnadareski (approved):

PR #980, opened 2026-06-18
Enumerate /dev/sr* optical drives on Linux

gmipf:
On Linux, `DriveInfo.GetDrives()` does not surface unmounted optical block
devices, so optical drives are missing from the drive list unless a disc is
already mounted. This enumerates `/dev/sr*` directly so discs can be dumped
without mounting first.

### What this PR does
- After the existing `GetDriveList()` logic, on Unix, scans `/dev` for kernel
  optical device nodes (`sr` followed by digits: `sr0`, `sr1`, …) and adds any
  not already present as optical `Drive` entries.
- Adds `Drive.EnumerateUnixOpticalDevicePaths(string)` (the scan/match helper),
  covered by unit tests.

### What this PR doesn't do
- Does not change Windows or macOS enumeration. The branch is gated on
  `PlatformID.Unix`; macOS/BSD have no `/dev/sr*` nodes, so it is a no-op there.
- Does not wire the enumerated path into filesystem-based heuristics.
  `Drive.Name` is set to the device path (e.g. `/dev/sr0`), which is correct as
  the dumper argument but is not a mountable filesystem path, so volume-label /
  PS3-PS4-PS5 detection / protection scanning remain unaddressed for these
  entries. That is a larger change and intentionally out of scope here.
- No new dependencies.

### Tests
`MPF.Frontend.Test/DriveTests.cs` — 3 cases for the matcher: null/empty root,
missing directory, and that only `sr` + digits match (`sr0`, `sr1`, `sr15`)
while `src`, `srm`, `serial0`, `sr`, `sr0a` are excluded. The matcher runs
against a temp directory, so the test is platform-independent.

Built net20 → net10.0, all green; full MPF.Frontend.Test suite passes (505).

Relates to the existing `// TODO: Reduce reliance on DriveInfo` note in
`Drive.cs`. Happy to rebase or adjust scope.

Prepared with AI assistance (Claude Opus 4.8) and reviewed before submission.

### DIFF
--- a/MPF.Frontend.Test/DriveTests.cs
+++ b/MPF.Frontend.Test/DriveTests.cs
@@ -55,5 +55,54 @@ namespace MPF.Frontend.Test
         }
 
         #endregion
+
+        #region EnumerateUnixOpticalDevicePaths
+
+        [Fact]
+        public void EnumerateUnixOpticalDevicePaths_NullOrEmpty_ReturnsEmpty()
+        {
+            Assert.Empty(Drive.EnumerateUnixOpticalDevicePaths(null!));
+            Assert.Empty(Drive.EnumerateUnixOpticalDevicePaths(string.Empty));
+        }
+
+        [Fact]
+        public void EnumerateUnixOpticalDevicePaths_MissingDirectory_ReturnsEmpty()
+        {
+            string missing = Path.Combine(Path.GetTempPath(), "mpf-test-missing-" + Guid.NewGuid().ToString("N"));
+            Assert.Empty(Drive.EnumerateUnixOpticalDevicePaths(missing));
+        }
+
+        [Fact]
+        public void EnumerateUnixOpticalDevicePaths_OnlyMatchesSrFollowedByDigits()
+        {
+            string root = Path.Combine(Path.GetTempPath(), "mpf-test-dev-" + Guid.NewGuid().ToString("N"));
+            Directory.CreateDirectory(root);
+            try
+            {
+                File.WriteAllText(Path.Combine(root, "sr0"), string.Empty);
+                File.WriteAllText(Path.Combine(root, "sr1"), string.Empty);
+                File.WriteAllText(Path.Combine(root, "sr15"), string.Empty);
+                File.WriteAllText(Path.Combine(root, "src"), string.Empty);
+                File.WriteAllText(Path.Combine(root, "srm"), string.Empty);
+                File.WriteAllText(Path.Combine(root, "serial0"), string.Empty);
+                File.WriteAllText(Path.Combine(root, "sr"), string.Empty);
+                File.WriteAllText(Path.Combine(root, "sr0a"), string.Empty);
+
+                var actual = Drive.EnumerateUnixOpticalDevicePaths(root);
+
+                var actualNames = new List<string>();
+                foreach (var p in actual)
+                    actualNames.Add(Path.GetFileName(p));
+                actualNames.Sort(StringComparer.Ordinal);
+
+                Assert.Equal(new[] { "sr0", "sr1", "sr15" }, actualNames);
+            }
+            finally
+            {
+                Directory.Delete(root, recursive: true);
+            }
+        }
+
+        #endregion
     }
 }
--- a/MPF.Frontend.Test/DriveTests.cs
+++ b/MPF.Frontend.Test/DriveTests.cs
@@ -56,24 +56,24 @@ namespace MPF.Frontend.Test
 
         #endregion
 
-        #region EnumerateUnixOpticalDevicePaths
+        #region EnumerateUnixOpticalBlockPaths
 
         [Fact]
-        public void EnumerateUnixOpticalDevicePaths_NullOrEmpty_ReturnsEmpty()
+        public void EnumerateUnixOpticalBlockPaths_NullOrEmpty_ReturnsEmpty()
         {
-            Assert.Empty(Drive.EnumerateUnixOpticalDevicePaths(null!));
-            Assert.Empty(Drive.EnumerateUnixOpticalDevicePaths(string.Empty));
+            Assert.Empty(Drive.EnumerateUnixOpticalBlockPaths(null!));
+            Assert.Empty(Drive.EnumerateUnixOpticalBlockPaths(string.Empty));
         }
 
         [Fact]
-        public void EnumerateUnixOpticalDevicePaths_MissingDirectory_ReturnsEmpty()
+        public void EnumerateUnixOpticalBlockPaths_MissingDirectory_ReturnsEmpty()
         {
             string missing = Path.Combine(Path.GetTempPath(), "mpf-test-missing-" + Guid.NewGuid().ToString("N"));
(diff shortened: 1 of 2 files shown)


MPF.Frontend.Test/DriveTests.cs
```diff
@@ -55,5 +55,54 @@ public void ToInternalDriveTypeTest(DriveType driveType, bool expectNull)
...
+                File.WriteAllText(Path.Combine(root, "srm"), string.Empty);
+                File.WriteAllText(Path.Combine(root, "serial0"), string.Empty);
+                File.WriteAllText(Path.Combine(root, "sr"), string.Empty);
+                File.WriteAllText(Path.Combine(root, "sr0a"), string.Empty);
+
+                var actual = Drive.EnumerateUnixOpticalDevicePaths(root);
+
+                var actualNames = new List<string>();
+                foreach (var p in actual)
+                    actualNames.Add(Path.GetFileName(p));
```
mnadareski, line 96: All loops need to be surrounded by curly braces, even single-line ones

MPF.Frontend/Drive.cs
```diff
@@ -341,9 +341,95 @@ private static List<Drive> GetDriveList(bool ignoreFixedDrives)
             }
 #endif
 
+            // On Linux, DriveInfo.GetDrives() returns mount points and typically
+            // does not surface unmounted optical block devices (/dev/sr*). Enumerate
+            // them directly so users can dump discs without first mounting them.
+            if (Environment.OSVersion.Platform == PlatformID.Unix)
```
mnadareski: This doesn't impact MacOS, correct?

mnadareski: This whole block can probably be an alternative path to the existing Windows-specific code above. Possibly a good design would be to have a set of if/else statements in this method and then each OS-specific drive array building code in private or internal methods.

MPF.Frontend/Drive.cs
```diff
@@ -341,9 +341,95 @@ private static List<Drive> GetDriveList(bool ignoreFixedDrives)
...
+                var extra = new List<Drive>();
+                foreach (var devicePath in EnumerateUnixOpticalDevicePaths("/dev"))
+                {
+                    if (existingNames.Contains(devicePath))
+                        continue;
+
+                    Drive? d;
+                    try
+                    {
+                        d = Create(Frontend.InternalDriveType.Optical, devicePath);
```
mnadareski: The gated `.Add` statement can safely be put inside of this block, which could simplify how all of this is being handled anyway.

MPF.Frontend/Drive.cs
```diff
@@ -341,9 +341,95 @@ private static List<Drive> GetDriveList(bool ignoreFixedDrives)
...
+        internal static List<string> EnumerateUnixOpticalDevicePaths(string devRoot)
+        {
+            var result = new List<string>();
+            if (string.IsNullOrEmpty(devRoot) || !Directory.Exists(devRoot))
+                return result;
+
+            string[] candidates;
+            try
+            {
+                candidates = Directory.GetFiles(devRoot, "sr*");
```
mnadareski, line 426: Is it worth including `sg*` as well? I know some users prefer using those instead of `sr*`

gmipf:
Thanks for the detailed review. Addressed in 06f9c64f:

- Refactored `GetDriveList` into an OS branch with per-OS helpers (`MarkWindowsFloppyDrives`, `AppendUnixOpticalDrives`).
- Simplified the optical-device loop (the add is now inside the `try`).
- Added braces to the single-line loop in the test.

On the two open questions:

- **macOS**: correct, this is currently a no-op there. The branch is gated on `PlatformID.Unix`, but only `/dev/sr*` is scanned and those nodes don't exist on macOS. A proper equivalent is more involved: `/Volumes` only lists *mounted* discs (so it wouldn't cover the unmounted-drive case this fixes), and `/dev/disk*` isn't optical-specific. Reliable optical detection on macOS needs IOKit/DiskArbitration (or parsing `diskutil`). The OS-branch structure leaves a clean slot for a macOS branch — happy to add it as a follow-up, ideally with someone who can verify on real hardware since I can't.
- **`sg*`**: I'd lean toward keeping just `sr*` here. `sg` is the generic SCSI passthrough layer and isn't optical-specific (it also enumerates disks, tapes, etc.), so a bare `sg*` scan would surface non-optical devices. Picking the right `sg` node for an optical drive would need a sysfs correlation (`/sys/block/srN/device` → matching `sgN`), which feels like a separate change. Open to adding it if you'd prefer.

MPF.Frontend/Drive.cs
```diff
@@ -340,49 +354,50 @@ private static List<Drive> GetDriveList(bool ignoreFixedDrives)
...
-            if (Environment.OSVersion.Platform == PlatformID.Unix)
+        /// <summary>
+        /// Append Linux optical block devices (/dev/sr*) that DriveInfo did not surface.
+        /// DriveInfo.GetDrives() returns mount points and typically does not list unmounted
+        /// optical drives, so they are enumerated directly to let users dump discs without
+        /// mounting them first.
+        /// </summary>
+        /// <param name="drives">Drives already discovered via DriveInfo</param>
+        /// <returns>The drive array, extended with any optical devices not already present</returns>
+        private static Drive[] AppendUnixOpticalDrives(Drive[] drives)
```
mnadareski, line 369: This works, but I think I'm going to take a look at it after it gets merged to see if I can tweak it a little.

mnadareski:
For the future, copying the responses to the relevant individual comment threads makes it easier to track responses.

For the Mac bit, that's really weird and unfortunate how that's handled, but that can be a future PR. This one focusing on Linux is a great start.

For the `sg*` addition, I have a feeling it's going to highly depend on the users, but there is an implicit statement that users on Linux systems should at least be partially aware if `sr*` or `sg*` is more appropriate for their setup. So I'd like to have both, even if the `sg*` are listed after `sr*`.

MPF.Frontend/Drive.cs
gmipf: Sounds good — I'll list `sg*` after the `sr*` entries.

The right node is genuinely tool-dependent, which is itself a good argument for surfacing both:
- redumper opens with `O_RDWR | O_EXCL`; on a block device the kernel returns `EBUSY` for `O_EXCL` when the disc is mounted (desktops auto-mount data discs), so `sr*` can fail there. Its own drive list emits `sg*` for this reason.
- DiscImageCreator opens without `O_EXCL`, so either node works.
- Aaru dumps via either, but pulls device metadata from `/sys/block/<name>`, which only exists for `sr*` (the `sg*` nodes sit under `/sys/class/scsi_generic`), so `sr*` can give richer info.

The support lands in both the CLI and the Avalonia GUI. Only display caveat: the GUI drop-down is selection-only (no free-text), so with both node types listed a single physical drive appears twice — e.g. `sr0` and `sg0`. Distinct, labeled entries, and purely cosmetic; the underlying support is identical. The CLI takes a path argument directly, so it's unaffected.

mnadareski (approved):

PR #981, opened 2026-06-19
Mark Linux optical drives as active

gmipf:
Linux optical `/dev` nodes are never mount points, so `DriveInfo` always reports them as not-ready — they were enumerated but never marked active. This marks all `/dev/sr*` and `/dev/sg*` nodes active on enumeration, matching how Windows surfaces optical drives.

Per the discussion on this PR, this drops the `CDROM_DRIVE_STATUS` ioctl approach in favor of always-active to avoid extra system calls. Whether a disc is actually loaded is left to the dumping program. Rebased onto current master, so the change is a small addition in `AppendUnixOpticalDrives`.

### DIFF
--- a/MPF.Frontend/Drive.cs
+++ b/MPF.Frontend/Drive.cs
@@ -391,7 +391,14 @@ namespace MPF.Frontend
                 {
                     var d = Create(Frontend.InternalDriveType.Optical, devicePath);
                     if (d != null)
+                    {
+                        // A Linux optical /dev node is never a mount point, so DriveInfo always
+                        // reports it as not-ready. Mark it active so it matches how Windows
+                        // surfaces optical drives; whether a disc is actually loaded is left to
+                        // the dumping program rather than probing the device here.
+                        d.MarkedActive = true;
                         extra.Add(d);
+                    }
                 }
                 catch
                 {


MPF.Frontend/Drive.cs
```diff
@@ -531,5 +545,333 @@ private static bool HasDeviceIndexSuffix(string name, string prefix)
...
+            finally
+            {
+                if (fd >= 0)
+                    NativeClose(fd);
+            }
+        }
+
+        #region ISO 9660 Filesystem (read directly from the block device)
+
+        private const string Iso9660DriveFormat = "CDFS";              // mirror DriveInfo.DriveFormat on Windows
```
mnadareski: General comment as I look at the logic here: Constants, fields, and properties should all be at the top of the file. You can have them in regions there, but they just need to be at the top.

mnadareski: For ISO stuff, there is a lot of support code in SabreTools.Serialization, but that version of the library is not yet included in MPF due to a chain of dependencies. If this can hold on until then, there are many constants and the like that would be at your disposal.

gmipf: Done — moved the native-interop constants into a `Constants` region at the top of the class.

gmipf: Sounds good — I've split the ISO reading out of this PR, so it now covers just the disc-present status (`MarkedActive`). The label/size reading can come back as a follow-up built on `SabreTools.Serialization` once that's in MPF, rather than the hand-rolled descriptor parser. Thanks for the pointer.

gmipf:
On the Windows worry: none of these tests touch a real device or `/dev`. Each one builds a temp directory tree and passes the `dev`/sysfs roots in as parameters, so they need no optical drive — and nothing in them uses a Linux-only file API (no symlinks; the expected paths are built with `Path.Combine`, not literal `/dev/...` strings). So they should be portable, though I can only run them on Linux myself.

The three that model the Linux `/dev` + sysfs layout are `EnumerateUnixOpticalBlockPaths`, `EnumerateUnixOpticalGenericPaths`, and `ResolveUnixOpticalBlockNode`; `MapCdromStatusToMediaPresent` is plain status-to-bool logic. If you'd rather make the intent explicit, I'm happy to gate just those three to Linux and leave the status-mapping test running everywhere — whichever you prefer. No rush on the rest.

mnadareski:
At the risk of coming at this from too broad and basic of an angle, I think it may be okay to always just set Linux drives as active. There's a certain level of expectation on users on Linux to even know what those devices are, and just having all of them "available" is fine for now. 

In my eyes, the bigger issue would be the mounted path for things like information extraction and protection scanning. I don't really know of a good way to expose that in the UI. It's already in CLI as a separate option.

gmipf:
Good point on the mounted path — that's the harder problem, and I'd keep it separate. The filesystem-dependent bits (information extraction, protection scanning) need the mount root rather than the device node, so they don't really belong with the active flag.

On the active flag itself: `MarkedActive` is already non-blocking in MPF. Every drive stays listed and selectable, and dumping an inactive one still works — it just hits the existing "No Disc Detected — continue?" confirmation and proceeds on Yes. On Windows it's already derived from real media state (`DriveInfo.IsReady`), so this PR only brings Linux to parity rather than adding any gating. Detecting (vs. always-active) is what gives the at-a-glance "has a disc" indication and lets auto-select land on the loaded drive; always-active would lose both and diverge from Windows.

That said, if you'd prefer the simpler always-active path for now, I'm happy to switch it — your call.

gmipf:
One thing I initially missed on my end: on Windows the drive list already includes fixed (and removable) drives when `IgnoreFixedDrives` is off. The Linux branch I added only appends optical nodes, so it's currently more restrictive than Windows — fixed disks never show up regardless of that option.

If it'd be welcome, I could look into bringing Linux to parity as a separate follow-up: having the Linux enumeration honor that same `IgnoreFixedDrives` option, reusing the existing toggle rather than adding any new UI. Optical stays the default; this would just stop Linux from being the odd one out. Not part of this PR — just wanted to check whether that direction is something you'd want before I dig into it.

mnadareski:
As a followup to both comments above:

- Yes, I will strongly prefer just setting all `/dev/sg*` and `/dev/sr*` to being active so we can avoid more system calls for now. If this changes in the future, we can try to make it more accurate to how the dumping programs deal with active drives.
- A future change should be made to allow for a separate mounted drive path for Linux and Mac, similar to how CLI and Check require. This could either be a fully-manual step or one that uses some OS-specific code to handle internally.
- Fixed drives and floppy drives were only added to the main UI because of their support in Aaru and DiscImageCreator. Support for those are not required in Linux or Mac and may be removed as an option in future core code even on Windows.

gmipf:
Switched this to the always-active approach you preferred: all enumerated `/dev/sr*` and `/dev/sg*` are marked active on enumeration, dropping the `CDROM_DRIVE_STATUS` ioctl entirely — no extra system calls. Rebased onto current master, so it's now a small addition in `AppendUnixOpticalDrives`. Disc presence is left to the dumping program, as discussed.

mnadareski (approved):

PR #990, opened 2026-06-23
Enumerate fixed/removable drives on Linux and order the drive list

gmipf:
Two related improvements to the Linux drive list, building on #980 (sr/sg enumeration) and #981 (optical drives active).

## What this PR does

**1. Enumerate fixed and removable drives on Linux (gated)**
On Windows the drive list includes fixed and removable drives when `IgnoreFixedDrives` is off; the Linux branch only added optical nodes, so it was more restrictive. This adds `AppendUnixFixedDrives`, which enumerates whole block devices from `/sys/block` (mirroring the optical enumerator) and honors the same existing `Options.GUI.IgnoreFixedDrives` toggle — no new option or UI.
- Reads world-readable sysfs; the `removable` flag selects `HardDisk` vs `Removable`, size comes from sysfs.
- The fixed-drive enumerator skips non-target nodes: virtual block devices (`dm-`, `loop`, `md`, `nbd`, `ram`, `zd`, `zram`), floppies (`fd`), and optical drives (`sr`, which are real targets but surfaced separately by the optical enumerator).
- Fixed disks are marked active; removable drives are active only when media is present (mirrors `DriveInfo.IsReady` on Windows).
- On Unix, fixed/removable come from sysfs rather than `DriveInfo` (which only exposes mount points like `/`, not dumpable device nodes). The Windows path is unchanged.

**2. Order the drive list**
Sorts the list with `SabreTools.Text.Compare.NaturalComparer` so device indices order numerically (e.g. `sr2` before `sr10`) instead of lexically; Windows drive letters keep a natural, digit-aware order too.

The Linux enumeration helpers live in a separate partial class (`Drive.Linux.cs`), following the existing `ThemeService.Linux.cs` pattern, so a future macOS implementation can add its own counterpart.

## What this PR does not do
- No new UI or options (reuses `IgnoreFixedDrives`).
- Does not grant non-root raw device access — enumeration only reads sysfs; dumping a fixed disk still needs the usual privileges.
- Does not touch macOS/BSD enumeration, and does not add the mounted-path/filesystem reader for protection scans (separate future work).
- No dependencies added.

## Tests
`MPF.Frontend.Test/DriveTests.cs`: hermetic tests (temp fake-sysfs trees, injected roots) for the optical and fixed/removable enumerators (sr/sg matching, excluded-prefix coverage, removable/size classification, missing-metadata fallback). No real devices touched. Built net20 through net10.0, all green.


MPF.Frontend/Drive.cs
```diff
@@ -316,8 +316,12 @@ private static List<Drive> GetDriveList(bool ignoreFixedDrives)
             // Apply OS-specific drive adjustments
             if (Environment.OSVersion.Platform == PlatformID.Win32NT)
                 MarkWindowsFloppyDrives(drives);
-            else if (Environment.OSVersion.Platform == PlatformID.Unix)
+            else if (isUnix)
```
mnadareski, line 323: Since this is now multiline, the `if` statement above should also have curly braces.

MPF.Frontend/Drive.cs
```diff
@@ -519,6 +523,305 @@ private static bool HasDeviceIndexSuffix(string name, string prefix)
...
+            return result;
+        }
+
+        /// <summary>
+        /// sysfs block-device name prefixes that are not user-facing dump targets: loopback
+        /// images, ramdisks, device-mapper and software RAID virtual devices, floppy and
+        /// network block devices, ZFS volumes, and optical drives (surfaced separately via
+        /// the optical enumerator).
+        /// </summary>
+        private static readonly string[] _excludedUnixBlockPrefixes =
```
mnadareski: Not sure why this is bugging me, but if you can have these be ordered alphabetically, I'd appreciate it. If you want to include what each one is from the description above as a trailing inline comment, that's good as well.

MPF.Frontend/Drive.cs
```diff
@@ -519,6 +523,305 @@ private static bool HasDeviceIndexSuffix(string name, string prefix)
...
+                // Unreadable size; report it as unknown
+            }
+
+            return 0;
+        }
+
+        /// <summary>
+        /// A Linux fixed or removable block device discovered via sysfs.
+        /// </summary>
+        internal sealed class UnixBlockDevice
```
mnadareski: Please extract this class into a new file since it's not `private`.

MPF.Frontend/Drive.cs
```diff
@@ -519,6 +523,305 @@ private static bool HasDeviceIndexSuffix(string name, string prefix)
...
+
+        /// <summary>
+        /// Compare two strings with runs of digits treated as numbers, so "sr2" sorts before
+        /// "sr10" (and multi-digit indices order correctly). Non-digit characters are compared
+        /// ordinally.
+        /// </summary>
+        /// <param name="a">First string</param>
+        /// <param name="b">Second string</param>
+        /// <returns>Negative, zero, or positive per IComparer conventions</returns>
+        internal static int NaturalCompare(string a, string b)
```
mnadareski: I'm fairly certain that this is a copy of the existing `SabreTools.Text.Compare` code for some reason. Please use `SabreTools.Text.Compare.NaturalComparer` instead.

MPF.Frontend/Drive.cs
```diff
@@ -519,6 +523,305 @@ private static bool HasDeviceIndexSuffix(string name, string prefix)
...
+        /// Append Linux fixed and removable block devices that DriveInfo did not surface.
+        /// On Unix, DriveInfo.GetDrives() only reports mount points (e.g. "/"), not the raw
+        /// device nodes (/dev/sda, /dev/nvme0n1, ...) needed for dumping, so whole-disk
+        /// block devices are enumerated directly from sysfs instead. Enumeration relies only
+        /// on world-readable sysfs and does not require elevated privileges; only the later
+        /// raw read during a dump does.
+        /// </summary>
+        /// <param name="drives">Drives already discovered via DriveInfo</param>
+        /// <returns>The drive array, extended with any block devices not already present</returns>
+        private static Drive[] AppendUnixFixedDrives(Drive[] drives)
```
mnadareski: This should probably have an OS check in it just in case. Same goes for the other Linux-only helpers.

mnadareski: For this and other Linux-specific helpers, I think it will end up cleaner to put them in a new partial `Drive.Unix.cs` or `Drive.Linux.cs` class. This will also help when the time comes to add Mac, which would need its own partial class as well.

mnadareski (commented):
Some general comments before I can check into the logic.

MPF.Frontend/Drive.cs
gmipf: Done — braced the `if (… == PlatformID.Win32NT)` branch so the whole if/else chain is consistent.

MPF.Frontend/Drive.cs
gmipf: Done — switched to `SabreTools.Text.Compare.NaturalComparer` (available via `SabreTools.IO`) and dropped the hand-rolled comparer. One behavior note: this also removes the optical-first ranking, so the list is now plain natural order — `/dev/sg*` sorts before `/dev/sr*`. If you'd rather keep optical block nodes first, I can put a small rank ahead of the comparer; otherwise it stays plain natural.

MPF.Frontend/Drive.cs
gmipf: Done — added an `Environment.OSVersion.Platform != PlatformID.Unix` guard to `AppendUnixOpticalDrives` and `AppendUnixFixedDrives` (using `PlatformID.Unix` rather than `OperatingSystem.IsLinux()` since the project targets down to net20). I left it off the `Enumerate*` helpers, since those take injected roots and are covered by OS-independent unit tests — a platform check there would tie those tests to Linux. Happy to add it there too if you'd prefer.

gmipf: Done — moved the Linux helpers (and the exclude-prefix field) into a new partial `MPF.Frontend/Drive.Linux.cs`, following the `ThemeService.cs` / `ThemeService.Linux.cs` pattern. A future macOS version can add `Drive.MacOS.cs`.

MPF.Frontend/Drive.cs
gmipf: Done — reordered the prefixes alphabetically and added a trailing comment describing each.

MPF.Frontend/Drive.cs
gmipf: Done — moved it to its own file, `MPF.Frontend/UnixBlockDevice.cs`.

MPF.Frontend/Drive.cs
mnadareski: Natural order only is perfectly fine. Honestly, I think a lot of users use `sg*` because of prior instruction.

mnadareski (approved):

PR #998, opened 2026-06-27
Add live tool output console for the Linux GUI

gmipf:
## What

On Linux the dumping tools (redumper, DiscImageCreator, Aaru) run without a console window of their own, so their live progress output was not visible anywhere in the GUI. This adds an in-app **Program Output** window that streams the tool's stdout/stderr live while it runs.

## Scope: Linux only, for now

The console is wired up **only on Linux**. On Windows, macOS, the WPF GUI and the CLI nothing changes — the output sink stays `null`, so the original `UseShellExecute` launch path is used exactly as before. I kept this conservative for the first cut; if you're happy with the approach, a natural follow-up would be to unify it across all platforms (and drop the separate CMD window on Windows). Happy to discuss.

## How it works

- **MPF.ExecutionContexts** – `BaseExecutionContext` gains an optional `Action<string>? OutputReceived`. When set, the tool is started with `RedirectStandardOutput/Error` and read raw (char-level, carriage-return aware) on two background threads, so `\r` progress redraws don't flood the log. When `null` (every current caller), the launch path is unchanged. Kept net20-safe (no Channels/async) for the multi-targeted projects.
- **MPF.Frontend** – `DumpEnvironment.ToolOutputReceived` forwards the stream. A minimal `IToolOutputConsole` seam (Open / Append / NotifyToolExited) lets a frontend present the output without the frontend layer taking a UI dependency. New `GuiSettings.ToolConsoleAutoClose` option, round-tripped through `OptionsLoader`.
- **MPF.Avalonia** – `ToolOutputWindow` renders the stream in a virtualized, terminal-style view (fixed-height rows, `\r`/`\n` line discipline, copy / select-all). Shown **modeless and owned** beside the main window, so the main window and its Stop button stay fully responsive. An "auto-close when the program exits" checkbox (default on) is persisted to the config.

## Localization

New user-facing strings added to `Strings.xaml` and `Strings.de.xaml`.

## Testing

- Full solution builds clean across all target frameworks (net20 → net10.0): 0 warnings, 0 errors.
- `MPF.ExecutionContexts.Test` (615) and `MPF.Frontend.Test` (529) pass with 0 failures on net8.0/net9.0/net10.0.
- Manually verified on Fedora KDE (Avalonia, .NET 10) with real redumper and DiscImageCreator dumps of a PhotoCD disc: live output streams correctly, the window stays responsive, Stop works, and the auto-close preference persists across restarts.

---

### DIFF
--- a/CHANGELIST.md
+++ b/CHANGELIST.md
@@ -1,3 +1,7 @@
+### WIP (xxxx-xx-xx)
+
+- Add live tool output console for the Linux GUI (gmipf)
+
 ### 3.8.2 (2026-07-01)
 
 - Output used configuration path for commandline programs
--- a/MPF.Avalonia/Helpers/OutputPump.cs
+++ b/MPF.Avalonia/Helpers/OutputPump.cs
@@ -1,0 +1,121 @@
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Channels;
+using System.Threading.Tasks;
+
+namespace MPF.Avalonia.Helpers
+{
+    /// <summary>
+    /// Decouples the process-output producer from the (slow) UI consumer with a
+    /// bounded channel, so a firehose of output can never block the reader thread
+    /// or grow memory without bound -- the two failure modes behind the old lag.
+    ///
+    /// - Producer side (<see cref="Post"/>) is called from the stdout/stderr read
+    ///   loops. It never blocks: when the consumer falls behind and the buffer is
+    ///   full, the OLDEST queued chunk is dropped (<see cref="BoundedChannelFullMode.DropOldest"/>).
+    ///   The raw stream is still written to the log file in full elsewhere, so
+    ///   dropping only affects the live on-screen tail, never the saved log.
+    /// - Consumer side drains everything currently available in ONE batch
+    ///   (<see cref="DrainAvailable"/>), so the UI does a single dispatch per
+    ///   flush tick instead of one per line. That is what keeps a high output
+    ///   rate from saturating the UI thread.
+    /// </summary>
+    public sealed class OutputPump
+    {
+        private readonly Channel<string> _channel;
+
+        public OutputPump(int capacity = 8192)
+        {
+            _channel = Channel.CreateBounded<string>(new BoundedChannelOptions(capacity)
+            {
+                FullMode = BoundedChannelFullMode.DropOldest,
+                SingleReader = true,
+                SingleWriter = false,
+            });
+        }
+
+        /// <summary>
+        /// Producer: enqueue a raw output chunk. Never blocks, never throws on full.
+        /// </summary>
+        public void Post(string chunk)
+        {
+            // With DropOldest, TryWrite always succeeds for a bounded channel and
+            // silently evicts the oldest item when full -- exactly the backpressure
+            // we want for a live tail. After completion TryWrite returns false, which
+            // we ignore (output arriving after the window closed is simply dropped).
+            _channel.Writer.TryWrite(chunk);
+        }
+
+        /// <summary>
+        /// Signal that no more output will arrive.
+        /// </summary>
+        public void Complete() => _channel.Writer.TryComplete();
+
+        /// <summary>
+        /// Drain every chunk currently queued into one list, without waiting.
+        /// Caller feeds the batch to a <see cref="TerminalBuffer"/> and pushes a
+        /// single UI update. <paramref name="max"/> caps a single drain so one
+        /// pathological burst cannot monopolize a flush tick.
+        /// </summary>
+        public List<string> DrainAvailable(int max = 4096)
+        {
+            var batch = new List<string>();
+            while (batch.Count < max && _channel.Reader.TryRead(out string? chunk))
+            {
+                batch.Add(chunk);
+            }
+
+            return batch;
+        }
+
+        /// <summary>
+        /// Await at least one chunk being available (or completion). For the flush loop.
+        /// </summary>
+        public ValueTask<bool> WaitToReadAsync(CancellationToken ct = default)
+            => _channel.Reader.WaitToReadAsync(ct);
+
+        /// <summary>
+        /// Reference flush loop: wait, drain a batch, feed the buffer, raise one
+        /// onBatch callback (with the number of chunks consumed in the batch), then
+        /// throttle. In the GUI this callback marshals to the UI thread once per
+        /// batch. Throttling is what bounds the dispatch rate.
+        /// </summary>
+        public async Task RunAsync(TerminalBuffer buffer, Action<int> onBatch, int flushIntervalMs = 75, CancellationToken ct = default)
+        {
+            try
+            {
+                while (await WaitToReadAsync(ct).ConfigureAwait(false))
+                {
+                    List<string> batch = DrainAvailable();
+                    if (batch.Count > 0)
+                    {
+                        foreach (string chunk in batch)
+                        {
+                            buffer.Feed(chunk);
+                        }
+
+                        onBatch(batch.Count);
+                    }
+
+                    if (flushIntervalMs > 0)
+                        await Task.Delay(flushIntervalMs, ct).ConfigureAwait(false);
+                }
+            }
+            catch (OperationCanceledException)
+            {
+                // expected on shutdown
+            }
+
+            // Final drain after completion so the tail is not lost.
+            List<string> tail = DrainAvailable();
+            foreach (string chunk in tail)
+            {
+                buffer.Feed(chunk);
+            }
+
+            buffer.Flush();
+            onBatch(tail.Count);
+        }
+    }
+}
--- a/MPF.Avalonia/Helpers/TerminalBuffer.cs
+++ b/MPF.Avalonia/Helpers/TerminalBuffer.cs
@@ -1,0 +1,237 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace MPF.Avalonia.Helpers
+{
+    /// <summary>
+    /// Minimal terminal line-discipline for redirected child-process output.
+    ///
+    /// The point: read RAW characters from stdout (never <c>ReadLine</c>) and
+    /// interpret carriage returns the way a real terminal does. Dump tools
+    /// (redumper, DiscImageCreator, Aaru) update their progress with a bare
+    /// <c>\r</c> (no <c>\n</c>) so the same physical line is redrawn in place.
+    /// <c>ReadLine</c>/<c>OutputDataReceived</c> treat <c>\r</c> as a line break,
+    /// which explodes one updating line into thousands of separate lines -- the
+    /// spam + lag the maintainer saw when output was redirected years ago.
+    ///
+    /// Here, <c>\r</c> moves the write cursor to column 0 of the live line
+    /// (overwrite, not new line); <c>\n</c> commits the live line. Progress
+    /// updates therefore mutate a single live line and never grow the log.
+    ///
+    /// No UI dependency: pure and unit-testable. The UI binds the committed lines
+    /// to a virtualized list and the live line to one trailing row.
+    /// </summary>
+    public sealed class TerminalBuffer
+    {
+        private readonly StringBuilder _current = new();
+
+        /// <summary>
+        /// Write column within <see cref="_current"/>
+        /// </summary>
+        private int _cursor;
+
+        /// <summary>
+        /// Mid ANSI escape sequence (survives chunk boundaries)
+        /// </summary>
+        private bool _inEscape;
+
+        private readonly StringBuilder _escape = new();
+
+        /// <summary>
+        /// Lines committed so far (each was terminated by <c>\n</c>).
+        /// </summary>
+        public List<string> CommittedLines { get; } = new();
+
+        /// <summary>
+        /// The live line being assembled after the last <c>\n</c>.
+        /// </summary>
+        public string CurrentLine => _current.ToString();
+
+        /// <summary>
+        /// Raised once per committed line.
+        /// </summary>
+        public event Action<string>? LineCommitted;
+
+        /// <summary>
+        /// Raised when the live line changes (printable char, <c>\r</c> overwrite, erase).
+        /// </summary>
+        public event Action<string>? CurrentLineChanged;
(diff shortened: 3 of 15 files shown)


CHANGELIST.md
```diff
@@ -4,6 +4,7 @@
 - Resolve tilde-prefixed paths for binaries
 - Add feature to list path macros
 - Avoid possible unwanted filtering behavior
+- Add live tool output console for the Linux GUI (gmipf)
```
mnadareski: This will need to go into a new WIP section since there was a release after this PR was opened.

MPF.Avalonia/Helpers/OutputPump.cs
```diff
@@ -0,0 +1,106 @@
...
+        {
+            _channel = Channel.CreateBounded<string>(new BoundedChannelOptions(capacity)
+            {
+                FullMode = BoundedChannelFullMode.DropOldest,
+                SingleReader = true,
+                SingleWriter = false,
+            });
+        }
+
+        /// <summary>Producer: enqueue a raw output chunk. Never blocks, never throws on full.</summary>
```
mnadareski: Even short summaries should use multiple lines. Remarks can be single-line, though.

MPF.Avalonia/Helpers/OutputPump.cs
```diff
@@ -0,0 +1,106 @@
...
+        /// <summary>
+        /// Drain every chunk currently queued into one list, without waiting.
+        /// Caller feeds the batch to a <see cref="TerminalBuffer"/> and pushes a
+        /// single UI update. <paramref name="max"/> caps a single drain so one
+        /// pathological burst cannot monopolize a flush tick.
+        /// </summary>
+        public List<string> DrainAvailable(int max = 4096)
+        {
+            var batch = new List<string>();
+            while (batch.Count < max && _channel.Reader.TryRead(out string? chunk))
```
mnadareski, line 64: Single-line `while`, `for`, and `foreach` loops still require curly braces.

MPF.Avalonia/Helpers/TerminalBuffer.cs
```diff
@@ -0,0 +1,210 @@
...
+    /// (overwrite, not new line); <c>\n</c> commits the live line. Progress
+    /// updates therefore mutate a single live line and never grow the log.
+    ///
+    /// No UI dependency: pure and unit-testable. The UI binds the committed lines
+    /// to a virtualized list and the live line to one trailing row.
+    /// </summary>
+    public sealed class TerminalBuffer
+    {
+        private readonly StringBuilder _current = new();
+        private int _cursor;            // write column within _current
```
mnadareski: The posfix comments could be made into summaries.

MPF.Avalonia/Helpers/TerminalBuffer.cs
```diff
@@ -0,0 +1,210 @@
...
+        public event Action<string>? LineCommitted;
+
+        /// <summary>Raised when the live line changes (printable char, <c>\r</c> overwrite, erase).</summary>
+        public event Action<string>? CurrentLineChanged;
+
+        /// <summary>
+        /// Feed a raw chunk of output. Chunk boundaries are arbitrary: a <c>\r\n</c>
+        /// or an escape sequence may be split across two calls and is handled.
+        /// </summary>
+        public void Feed(string chunk)
```
mnadareski: Can a `chunk` be `null`? If so, make sure that's explicit due to nullable contexts.

MPF.Avalonia/Helpers/TerminalBuffer.cs
```diff
@@ -0,0 +1,210 @@
...
+
+                    case '\t':
+                        do { WriteChar(' '); } while (_cursor % 8 != 0);
+                        liveChanged = true;
+                        break;
+
+                    case '\b':
+                        if (_cursor > 0)
+                            _cursor--;
+                        break;
```
mnadareski, line 105: I'm surprised this didn't yell with auto-formatting. As much as it's not really fun, if there's a block before `break`, there should be an extra newline before the `break` statement.

MPF.Avalonia/Helpers/TerminalBuffer.cs
```diff
@@ -0,0 +1,210 @@
...
+                CommitLine();
+        }
+
+        private void WriteChar(char c)
+        {
+            if (_cursor < _current.Length)
+                _current[_cursor] = c;
+            else
+                _current.Append(c);
+            _cursor++;
```
mnadareski, line 138: Same where with the extra newline being needed after the block.

MPF.Frontend/Options.cs
```diff
@@ -444,6 +445,12 @@ public class GuiSettings
         /// <remarks>Version 1 and greater</remarks>
         public bool OpenLogWindowAtStartup { get; set; } = true;
 
+        /// <summary>
+        /// Automatically close the separate tool-output window when the dumping program exits
+        /// </summary>
+        /// <remarks>Version 1 and greater; currently used by the Linux GUI tool-output window</remarks>
+        public bool ToolConsoleAutoClose { get; set; } = true;
```
mnadareski, line 452: Could this also be extended to the existing normal OS console windows as well? I think there's invocation stuff that allows it to remain open after completion, but they are probably set to `false`.

MPF.ExecutionContexts/BaseExecutionContext.cs
```diff
@@ -195,26 +209,82 @@ protected abstract void SetDefaultParameters(string? drivePath,
...
-            process.WaitForExit();
+
+            if (redirect)
+            {
+                // Pump stdout and stderr on dedicated threads using RAW reads (never
+                // ReadLine): dump tools redraw progress with a bare '\r', and ReadLine
+                // would treat every '\r' as a new line -- the exact spam/lag that made
+                // redirected output unusable before. The consumer interprets '\r'.
+                Thread outThread = StartStreamPump(process.StandardOutput);
+                Thread errThread = StartStreamPump(process.StandardError);
```
mnadareski, line 241: Do these need to be defined before the process is started?

MPF.Avalonia/Windows/MainWindow.axaml.cs
```diff
@@ -116,10 +116,18 @@ private void OnOpened(object? sender, EventArgs e)
             if (MainViewModel.Options.GUI.ShowDebugViewMenuItem)
                 DebugViewMenuItem!.IsVisible = true;
 
+            // On Linux the dumping tool has no console window of its own, so stream its
+            // live output into a separate MPF window. Windows/macOS keep the tool's own
+            // console (leave this null), preserving the original behavior.
+            IToolOutputConsole? toolConsole = OperatingSystem.IsLinux()
```
mnadareski, line 122: I don't recall, but does MacOS open a new terminal? Or are we just playing it safe right now and only enabling it for MacOS if needed?

MPF.Avalonia/Windows/ToolOutputWindow.axaml.cs
```diff
@@ -0,0 +1,267 @@
...
+        private readonly TerminalBuffer _buffer = new();
+        private readonly OutputPump _pump = new(capacity: 8192);
+        private readonly CancellationTokenSource _cts = new();
+        private readonly Task _pumpTask;
+        private int _committedIndex;            // committed lines already pushed to the UI
+
+        private bool _autoScroll = true;        // follow the tail; toggled only by user input
+        private bool _scrollQueued;             // coalesces deferred ScrollToEnd posts
+        private bool _scrollBarHooked;          // vertical scrollbar Scroll event subscribed
+        private ScrollViewer? _scroller;        // ListBox's inner scroll viewer (found lazily)
```
mnadareski: Postfix comments could be made into summaries.

CHANGELIST.md
gmipf: Done — rebased onto current master and moved this into a fresh `### WIP (xxxx-xx-xx)` section above the released `3.8.2`, so the shipped release entries stay intact.

MPF.Avalonia/Helpers/OutputPump.cs
gmipf: Fixed — expanded every single-line `<summary>` to the multi-line form across the new files (`OutputPump`, `TerminalBuffer`). Left the `<remarks>` blocks as they were.

MPF.Avalonia/Helpers/OutputPump.cs
gmipf: Braced — added curly braces to every single-line `while`/`for`/`foreach` in the new code: this `DrainAvailable` loop plus the ones in `RunAsync`, `TerminalBuffer`, `ToolOutputWindow`, and `BaseExecutionContext`.

MPF.Avalonia/Helpers/TerminalBuffer.cs
gmipf: Done — converted the postfix comments on `_cursor` and `_inEscape` into multi-line `<summary>` docs.

MPF.Avalonia/Helpers/TerminalBuffer.cs
gmipf: It can. The reader threads only ever pass non-null strings today, but `Feed` is public and already defends against null/empty via the `IsNullOrEmpty` guard. I made that explicit by changing the parameter to `string?`, which matches the guard and keeps the nullable contract honest — `IsNullOrEmpty` narrows it back to non-null for the rest of the method, so there are no nullable warnings.

MPF.Avalonia/Helpers/TerminalBuffer.cs
gmipf: Added the blank line before `break` here (the `\b` case), and applied the same in `EraseInLine`'s `case 0` and `case 1`.

MPF.Avalonia/Helpers/TerminalBuffer.cs
gmipf: Fixed — blank line after the if/else block, before `_cursor++`.

MPF.Frontend/Options.cs
gmipf: Good question. Right now this flag only governs the Linux tool-output window, because that's the only path this PR redirects. On Windows/macOS the tool still launches via `UseShellExecute` and owns its own console, so whether that window stays open after the tool exits is decided by the console/shell (e.g. `cmd /k` vs `/c`), not by MPF today. Extending the setting to those would mean changing how the native consoles are launched — or redirecting them too — which is a platform-specific change beyond this PR's scope.

I scoped the remark to the Linux window for now to keep the meaning honest. Happy to look at unifying it as a follow-up if you'd like.

MPF.ExecutionContexts/BaseExecutionContext.cs
gmipf: They can't be, and don't need to be. `process.StandardOutput`/`StandardError` don't exist until after `Start()`, so the pump threads can only be created afterward — the same constraint the framework puts on `BeginOutputReadLine`, which also has to be called post-`Start()`.

The tiny window between `Start()` and the first `Read()` is covered by the OS pipe buffer, and because stdout and stderr each get their own draining thread, neither stream can wedge the other by filling its pipe. So the ordering here is correct and safe.

MPF.Avalonia/Windows/MainWindow.axaml.cs
gmipf: Playing it safe. I only have Linux to verify on. On macOS the tool launches via `UseShellExecute` from a windowed `.app`, and unlike Windows that generally does *not* pop a Terminal window — which means macOS users may well benefit from this same console. I just didn't want to flip behavior I can't test and risk regressing the Mac path.

Enabling it is a one-liner (`OperatingSystem.IsLinux()` → `OperatingSystem.IsLinux() || OperatingSystem.IsMacOS()`). Happy to add that if you can confirm macOS doesn't spawn its own terminal, or to keep it Linux-only for this PR and follow up separately.

MPF.Avalonia/Windows/ToolOutputWindow.axaml.cs
gmipf: Done — converted the postfix field comments (`_committedIndex`, `_autoScroll`, `_scrollQueued`, `_scrollBarHooked`, `_scroller`) into `<summary>` docs.

gmipf:
Thanks for the review. I've addressed all the points and force-pushed `c6fb4d24`:

- Rebased onto current master and put the changelog line in a fresh `### WIP` section above the released `3.8.2` (this also clears the merge conflict).
- Style: single-line `<summary>` blocks expanded to multi-line; postfix field comments turned into summaries; braces added to all single-line loops; blank lines added before `break`/after blocks.
- `TerminalBuffer.Feed` parameter is now `string?` to make the null-tolerant contract explicit.
- Replied inline on the three questions (native-console auto-close, pump-thread ordering, macOS scope).

Builds clean with `TreatWarningsAsErrors` across every target (net5.0–net10.0 for the Avalonia stack, net20–net10.0 for `MPF.ExecutionContexts`); GUI smoke-tested on Linux. The earlier inline comments will show as outdated because of the rebase.

MPF.Frontend/Options.cs
mnadareski: Ah okay, so that wasn't one of the normal launch params. That's a bit unfortunate, but I guess it's a limitation when dealing with direct `Process` things in C#.

MPF.ExecutionContexts/BaseExecutionContext.cs
mnadareski: That makes sense. I wasn't sure the order of operations there as it's been a while since I had attempted output redirection.

MPF.Avalonia/Windows/MainWindow.axaml.cs
mnadareski: I'm good with playing it safe for now. MacOS will have its own challenges.

mnadareski:
Just for future reference: you don't need to do a force push to the branch to ensure a clean commit. I squash and merge PRs when they're set. Unfortunately, force pushes make it harder for me to validate any given comment I made, as not just the code I expected to see change is marked as changed.

mnadareski (approved):

PR #999, opened 2026-06-27
Fix language menu requiring two clicks in the Avalonia GUI

gmipf:
## What

In the Avalonia GUI the title-bar language menu required **two clicks** to switch language — the first click did nothing visible, the second applied it. The active-language check mark also never appeared.

## Cause

The WPF GUI's language `MenuItem`s use `IsCheckable="True"`, which toggles the item's `IsChecked` state on click *before* the `Click` handler runs. The shared `LanguageMenuItemClick` handler relies on that: it reads `IsChecked` to tell a fresh selection from re-clicking the already-active language (which it deliberately ignores, so the current language can't be deselected).

The Avalonia port of the menu dropped the equivalent setting — Avalonia's `ToggleType` — so clicking a language never toggled `IsChecked`. The handler's early-return branch (`if (!clickedItem.IsChecked) { clickedItem.IsChecked = true; return; }`) caught every initial click and returned; only the second click, with `IsChecked` now true, applied the language. With `ToggleType` at its default of `None`, Avalonia also renders no check glyph, so the active language was never marked.

## Fix

Add `ToggleType="CheckBox"` to the language items in `MainWindow.axaml`, mirroring the WPF `IsCheckable="True"`. This restores the toggle-before-Click behaviour the shared handler already expects, so language switches on the first click and the active item shows a check mark. No code-behind change was needed.

## Testing

- Full Avalonia project builds clean on net10.0 (0 warnings, 0 errors).
- Manually verified on Fedora KDE (Avalonia, .NET 10): switching between English and German now takes a single click each way, the dropdown header updates, and the check mark tracks the active language.

---


CHANGELIST.md
```diff
@@ -4,6 +4,7 @@
 - Resolve tilde-prefixed paths for binaries
 - Add feature to list path macros
 - Avoid possible unwanted filtering behavior
+- Fix language menu requiring two clicks in the Avalonia GUI (gmipf)
```
mnadareski: Same thing here as the other PR where this will need to go into a new WIP part of the changelog above.

MPF.Avalonia/Windows/MainWindow.axaml
```diff
@@ -70,19 +70,24 @@
 
         <StackPanel Grid.Column="2" Orientation="Horizontal" VerticalAlignment="Stretch">
           <Menu x:Name="LanguagesMenuBar" Height="26" VerticalAlignment="Center" Background="Transparent" FontSize="12">
+            <!-- ToggleType="CheckBox" mirrors the WPF GUI's IsCheckable="True" on these
```
mnadareski: That is incredibly weird but at least there's a way around it.

CHANGELIST.md
gmipf: Done. Merged `master` and moved this into the new WIP section above 3.8.2 — the four entries that had been in my WIP block already shipped in 3.8.2, so they are out of WIP now.

MPF.Avalonia/Windows/MainWindow.axaml
gmipf: For the merge: `master`'s Avalonia cleanup reindented this block, so it conflicted. I kept the `ToggleType="CheckBox"` items (and the explanatory comment) on top of the new indentation.

CHANGELIST.md
```diff
@@ -1,10 +1,26 @@
...
+- Enable new languages and regions
+- Update RedumpLib to 2.0.7
+- Enable new languages and regions
+- Fix media type dropdown display issue
+- Fix detection issue for floppy disks
+- Fix WPF issue with older .NET versions
+- Fix message when clipboard is disabled
+- Update RedumpLib to 2.0.8
+- Attempt to handle generic disc labels again
+- Add temporary hack for multiline Dat field output
```
mnadareski, line 23: Not 100% sure why this is showing as changed still. Did this do a clean merge with `main`?

mnadareski (approved):

CHANGELIST.md
gmipf: Clean 3-way merge (`git merge origin/master`) — the only conflicts were CHANGELIST.md and MainWindow.axaml, both resolved by hand. Line 23 (`Attempt to handle generic disc labels again`) shows as changed only because it was not on this branch before the merge: it landed on `master` after I forked (in e97d8081), and the PR is diffed against the merge base, so it reads as added even though `master` already has it. No duplicate was introduced — the merged CHANGELIST matches the ordering on `master`.

PR #1004, opened 2026-07-01
Add man page output for the frontends

gmipf:
### Summary

`MPF.CLI man` and `MPF.Check man` write a roff man page to standard output, rendered from the same command model that produces `--help`. Packagers currently hand-write these pages and re-sync them on every release; this removes that drift at the source.

This wires up the man page feature added to `SabreTools.CommandLine` (SabreTools/SabreTools.CommandLine#2). That release is out — 1.5.0 on nuget.org — and master already references it, so the draft note about the pin no longer applies and this is ready for review.

### What it does

- `MPF.CLI man` and `MPF.Check man` render a page from the existing feature/input model. No new plumbing: `man` is just another reference feature, like `Help`.
- The graphical frontend has no command-line model to render from, so a short hand-maintained page (`MPF.Avalonia/MPF.1`) is included for it.

A packager then does:

```
MPF.CLI man > MPF.CLI.1      # generate roff from the binary
# stamp the .TH date at build time
# install into the man path
```

Two intentional details:

- **The page is emitted before the configuration path and other startup output**, so the redirect above yields clean roff with no leading `Configuration path:` line. Generating it has no side effects: it never writes `config.json` and never runs first-run setup.
- **The `.TH` date is left blank** so a distributor stamps it at build time, keeping output reproducible. This is the only thing `mandoc -T lint` flags until it is stamped.

Installing the pages is left to distributors; this PR only adds the ability to produce them.

### Verification

Against the published `SabreTools.CommandLine` 1.5.0, on this branch with master merged in:

- The full solution builds clean across every target framework with `TreatWarningsAsErrors`, 0 warnings.
- Both pages generate with exit code 0 and empty standard error, from a runtime directory with no `config.json` — and none is created, neither there nor under `~/.config/mpf`.
- `groff -man -ww -z` reports no warnings for all three pages. `mandoc -T lint` is clean except the intended missing-date warning.
- The generated page picks up features added upstream since this branch was opened (`listpath`, for one) with no change here, which is the point of generating rather than hand-writing it.

### Choices made

Names in `.TH`/`NAME` follow the assembly names (`MPF.CLI`, `MPF.Check`, `MPF`), matching what is invoked; the GUI page sits next to its project; option descriptions use the verbose form so the page is self-contained. Each is a one-line change if you would rather have it otherwise.

---


gmipf:
`SabreTools.CommandLine` 1.5.0 is on nuget.org and master already references it, so the blocker for this draft is gone. Merged master in and marked this ready for review — no rebase, so your earlier view of the diff still lines up.

Verified against the published 1.5.0: the full target-framework sweep builds clean with no warnings, both pages generate with no side effects (no `config.json` written, in the runtime directory or under `~/.config/mpf`), and `groff -man -ww -z` is silent on all three pages. The generated page also picked up `listpath`, which was added upstream after this branch was opened — without a change here, which is the whole point of rendering it from the command model.

I dropped the open questions from the description and made the calls, so there is nothing to decide before merging; each is a one-line change if you want it different.

MPF.CLI/Program.cs
```diff
@@ -15,6 +15,18 @@ public class Program
     {
         public static void Main(string[] args)
         {
+            // Create the command set
+            var mainFeature = new MainFeature();
+            var commandSet = CreateCommands(mainFeature);
+
+            // The man page is rendered purely from the command model, so emit it
+            // before the configuration path and other startup output to keep the roff clean
```
mnadareski, line 23: How would the existing standard output impact the generated roff? Isn't the manpage output purely generated from the command set? The initial load attempt for the options is very intentionally done before processing anything else so the configuration file is created if it wasn't already.

MPF.Avalonia/MPF.1
```diff
@@ -0,0 +1,23 @@
+.\" Manual page for the MPF graphical frontend.
```
mnadareski, line 1: Why is there an Avalonia manpage being checked in as a part of this?

MPF.CLI/Program.cs
gmipf: The content is, yes — the roff is rendered purely from the command model. What the reorder protects is the stream it is rendered to: the page goes to standard output, and so does the startup output.

I measured it rather than argue it. I moved the `man` feature back into the switch next to `Help` — where it would sit without the reorder, i.e. after the load — and rebuilt.

On a machine that has no configuration yet (a package build, a container, CI), `MPF.CLI man` produces 97 bytes and exits 0:

```
Configuration path: 

First-run detected! Please verify the generated config.json and run again.
```

Not one line of roff. The first-run branch returns before any argument is looked at. A build step that redirects this into `MPF.CLI.1` ships an empty page, and the build stays green. (The path is blank there only because I ran with a fresh `HOME`; that part is incidental.)

With a configuration already present the page is emitted, but prefixed:

```
Configuration path: /home/<user>/.config/mpf/config.json

.\" Generated from the command-line model. Do not edit by hand.
.TH "MPF.CLI" "1" "" "3.8.3" "User Commands"
```

so the file no longer begins with a roff request.

`groff -man -ww -z` is silent on both of those. The check one would naturally put in a build catches neither.

With the reorder, the same command on the same pristine machine gives 1907 bytes starting at `.TH`, and groff is silent for the right reason.

The reorder skips the load only for `man`. Every other invocation still loads the options and creates the configuration before anything else, exactly as it does today — the ordering you describe is untouched for them.

MPF.Avalonia/MPF.1
gmipf: Because there is nothing to generate it from: the GUI takes no arguments, so it has no command set to render. The two command-line frontends generate their pages; this one has to be written by hand, and it is the shortest page that is still correct — name, synopsis, description, and a pointer to the other two.

It is in this PR because a distribution that installs an `MPF` binary is expected to ship `MPF.1` alongside it; without it, `man MPF` fails and packaging lint flags the omission.

If you would rather not carry a hand-maintained file in the tree, say so and I will drop it in a follow-up commit — the two generated pages stand on their own.

MPF.CLI/Program.cs
gmipf: One more thing that makes the scope of this smaller than it looks: `MPF.Check` needed no reordering at all. It has no startup output — `Main` goes straight to the command set, with no `LoadFromConfig` and no configuration path printed — so `man` sits in its switch next to `Help`, like every other standalone option, and its page comes out clean.

`MPF.CLI` is the only one that prints before it looks at an argument, so it is the only one that is reordered.

gmipf:
Merged master in (twice — it moved while I was working), no rebase, so your view of the diff still lines up. Only `CHANGELIST.md` conflicted; your entries keep their order and mine sit at the bottom of the WIP list.

Your `Add logical opposite flags to Check` from today is a good demonstration of why the page is rendered from the command model rather than maintained by hand: all six new flags — `--no-archives`, `--no-artifacts`, `--no-create-ird`, `--no-debug-protection`, `--no-delete`, `--no-placeholders` — are in the generated page without a single change on this branch. The `MPF.Check` page now documents 42 options, the `MPF.CLI` page 17.

Re-verified after the merge, on a machine with no configuration: both pages generate clean (`MPF.CLI` 1907 bytes, `MPF.Check` 3297 bytes, both starting at the roff comment), `groff -man -ww -z` is silent on both, and no `config.json` is written anywhere.

MPF.CLI/Program.cs
mnadareski: Okay, so the intent is that this command would be run in some sort of automated fashion that does not assume "normal" use of the program. I will keep that in mind when I'm looking at the code change.

MPF.Avalonia/MPF.1
mnadareski: In that case, why are there not pre-generated mangpages for Check and CLI? It feels like those could be checked in periodically. Or is there a concern that those would go out of date for automation too quickly?

gmipf: You're right that this is inconsistent, and my earlier answer argued past your point. Three frontends, two mechanisms: `MPF.1` is a file in the tree, the other two only exist if you run the binary. That is fair to object to.

I also want to retract the shape of my drift argument. It is not a reason to avoid checking pages in — it is a reason a checked-in page needs a guard. What it actually measures, since this branch was opened (2026-07-01):

- `eb239b60` (2026-07-02) removed `--username` and `--password`.
- `0fded51c` (2026-07-14) added ten flags, removed three, and fixed the `--timeout-secondss` typo.

A page checked in on 2026-07-01 and left alone would today be missing `--no-archives`, `--no-artifacts`, `--no-create-ird`, `--no-debug-protection`, `--no-delete`, `--no-json`, `--no-pull-all`, `--no-scan`, `--no-suffix`, `--no-zip`, and would still advertise `--username`, `--password`, `--disable-archives`, `--enable-debug`, `--include-artifacts` and `--timeout-secondss`. So "periodically" is the part I would make mechanical, not the part I would drop.

That is straightforward here: the output is byte-identical across runs, it is pure LF, and `check_pr.yml` already runs on `ubuntu-latest`. Checking in `MPF.Check.1` and `MPF.CLI.1` next to `MPF.1`, plus a step that regenerates and diffs them, gets the consistency you want and turns the drift above into a red build instead of a silent lie. It also helps the case this PR came from: a packager installs the files directly rather than running the binary mid-build — a generator that quietly falls back during a package build yields a green build and an empty man page, which is worse than shipping none.

One measured caveat before you pick. The page carries the version in its `.TH` line, so a checked-in copy is coupled to every release, not just to the interface. Building with `-p:VersionPrefix=9.9.9` changes exactly one line:

```
< .TH "MPF.CHECK" "1" "" "3.8.3" "User Commands"
> .TH "MPF.CHECK" "1" "" "9.9.9" "User Commands"
```

Over the last twelve months that is 12 `Bump version` commits against 26 commits touching `MPF.Check/Features/MainFeature.cs` — so roughly a third of the regenerations a checked-in page would demand would carry no option information at all, and between a bump and the next regeneration the page would state the wrong version. Generated at runtime the version is always the binary's own, which is why it is there today. If the pages go in the tree, I would drop the version from `.TH`, so the page changes if and only if the interface does.

On `MPF.1`, my earlier phrasing was sloppy and worth stating precisely: `MPF.Avalonia`'s `Main` forwards `args` straight to `StartWithClassicDesktopLifetime` and defines no command set of its own, so there is nothing to render into options. Its page is a name, a synopsis, a description and a pointer to the other two. Under the layout above it is a file in the tree like the others — only its source differs, because it has no option set to track.

Two corrections to my earlier comments, both against my own case:

1. I said the page "picked up `listpath`, which was added upstream after this branch was opened." Wrong. `c85e3d2d` is dated 2026-06-26 and is an ancestor of this branch's first commit (2026-07-01); it was already in the tree. `listpath` is in the page, but it was never evidence of drift.
2. I said `0fded51c` added "six new flags" and listed `--no-placeholders` among them. Wrong count and wrong list: it added ten, and `--no-placeholders` predates it.

Both came from reading diffs rather than generating the page. Everything above comes from generating it: at the current head, `MPF.Check` renders 42 options, `MPF.CLI` renders 17, and `groff -man -ww -z` is silent on both.

Tell me which you want and I will push it as a follow-up commit here: checked-in pages plus the CI check (with the version dropped from `.TH`), or the current runtime-only form.

mnadareski (approved):

PR #1005, opened 2026-07-01
Detect USB floppy drives on Linux and fix media type handling

gmipf:
## Summary

Reworks this PR into full USB floppy support for the Linux GUI, plus two related media-type fixes that the first change surfaced.

### Changes

**1. Detect USB floppy drives and always list floppies (`d3e61e7d`)**
USB floppy drives enumerate as removable SCSI disks (`/dev/sd*`), not `/dev/fd*` nodes, so a removable block device whose media matches a standard floppy geometry (360 KB to 2.88 MB) is now classified as `InternalDriveType.Floppy`. Floppy drives are listed regardless of the fixed-drive toggle, alongside optical drives, so a USB floppy appears without also surfacing fixed disks. Classifying the drive as `Floppy` makes `GetPhysicalMediaType` return `FloppyDisk`, which satisfies the removable/floppy check in `DumpEnvironment.ParametersValid`. This mirrors how Windows tags floppy media via `Win32_LogicalDisk.MediaType`.

**2. Fix media type dropdown not updating dumping parameters (`08bec53b`)**
The media type `ComboBox` in both the Avalonia and WPF frontends bound `SelectedItem` to a `CurrentMediaType` property that was renamed to `CurrentPhysicalMediaType` in the RedumpLib 1.11 migration (`a4c9ee6d`), which updated the code-behind but not the XAML markup. The stale binding failed silently, so changing the media type never wrote back to the view model and the generated DiscImageCreator command was never refreshed (it stayed `cd` even when Floppy Disk was selected). Both bindings now point at `CurrentPhysicalMediaType`.

**3. Reset detected media type when switching drives (`f2dcebf1`)**
`CacheCurrentMediaType` only set `_detectedPhysicalMediaType` on a successful detection and never cleared it, so a type detected for one drive (e.g. `FloppyDisk`) carried over when switching to a drive whose media could not be read — an empty optical drive kept showing Floppy Disk instead of the system default. The remembered type is now cleared at the start of each drive's caching, so only a fresh detection sets it.

### Testing

Tested end-to-end on Fedora with a NEC USB floppy drive (enumerated as `/dev/sdh`):
- The USB floppy is detected and listed alongside optical drives, with fixed disks hidden.
- The media type dropdown now switches the DiscImageCreator command live (`fd` and `cd`).
- Selecting an optical or other drive no longer sticks on Floppy Disk.
- An actual floppy dump was verified with Aaru (full read, 0 unreadable sectors, FAT12 detected).
- For what it's worth: DiscImageCreator's `fd` command segfaults on Linux for this drive (it reads the SCSI inquiry, then crashes before dumping), so the dump was done with Aaru. That crash is a DiscImageCreator issue, separate from this PR — noting it since the media type dropdown targets DIC.

`MPF.Frontend.Test`: 558/558 pass, clean build (merged current `master`).

### Note for packagers

Raw access to a USB floppy (a `root:disk` `/dev/sd*` block device) requires a udev rule granting the desktop user access, e.g.:

```
SUBSYSTEM=="block", ENV{ID_DRIVE_FLOPPY}=="1", GROUP="plugdev", MODE="0660"
```

This is a system-configuration matter outside MPF, but relevant if MPF is packaged for Linux.

### DIFF
--- a/MPF.Frontend/Drive.cs
+++ b/MPF.Frontend/Drive.cs
@@ -324,7 +324,10 @@ namespace MPF.Frontend
             {
                 drives = AppendUnixOpticalDrives(drives);
                 if (!ignoreFixedDrives)
+                {
                     drives = AppendUnixFixedDrives(drives);
+                    drives = AppendUnixFloppyDrives(drives);
+                }
             }
 
             return [.. drives];
--- a/MPF.Frontend/Drive.cs
+++ b/MPF.Frontend/Drive.cs
@@ -322,14 +322,18 @@ namespace MPF.Frontend
             }
             else if (isUnix)
             {
-                // Optical and floppy drives are removable dump targets, so surface them
-                // regardless of the fixed-drive toggle; only fixed and non-floppy removable
-                // disks are hidden unless the user opts to show them. Floppy runs before the
-                // fixed enumerator so a USB floppy is tagged Floppy and then skipped there.
+                // Optical drives are removable dump targets, so surface them regardless of the
+                // fixed-drive toggle. Floppy drives are hidden unless the user opts to show
+                // fixed drives, matching Windows (where floppies enumerate as removable media
+                // that is only queried alongside fixed drives) and because the default program,
+                // Redumper, does not support them. Floppy runs before the fixed enumerator so a
+                // USB floppy is tagged Floppy and then skipped there.
                 drives = AppendUnixOpticalDrives(drives);
-                drives = AppendUnixFloppyDrives(drives);
                 if (!ignoreFixedDrives)
+                {
+                    drives = AppendUnixFloppyDrives(drives);
                     drives = AppendUnixFixedDrives(drives);
+                }
             }
 
             return [.. drives];
--- a/MPF.Frontend/Drive.cs
+++ b/MPF.Frontend/Drive.cs
@@ -322,12 +322,14 @@ namespace MPF.Frontend
             }
             else if (isUnix)
             {
+                // Optical and floppy drives are removable dump targets, so surface them
+                // regardless of the fixed-drive toggle; only fixed and non-floppy removable
+                // disks are hidden unless the user opts to show them. Floppy runs before the
+                // fixed enumerator so a USB floppy is tagged Floppy and then skipped there.
                 drives = AppendUnixOpticalDrives(drives);
+                drives = AppendUnixFloppyDrives(drives);
                 if (!ignoreFixedDrives)
-                {
                     drives = AppendUnixFixedDrives(drives);
-                    drives = AppendUnixFloppyDrives(drives);
-                }
             }
 
             return [.. drives];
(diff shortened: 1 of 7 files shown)


MPF.Frontend/ViewModels/MainViewModel.cs
```diff
@@ -1217,6 +1217,12 @@ private void CacheCurrentMediaType()
             if (CurrentDrive is null)
                 return;
 
+            // Forget the type detected for any previously selected drive; only a fresh
```
mnadareski, line 1220: Oh, that actually makes a lot of sense. That probably explains a lot of weird interactions.

MPF.Frontend/Drive.cs
```diff
@@ -322,7 +322,12 @@ private static List<Drive> GetDriveList(bool ignoreFixedDrives)
             }
             else if (isUnix)
             {
+                // Optical and floppy drives are removable dump targets, so surface them
+                // regardless of the fixed-drive toggle; only fixed and non-floppy removable
+                // disks are hidden unless the user opts to show them. Floppy runs before the
+                // fixed enumerator so a USB floppy is tagged Floppy and then skipped there.
                 drives = AppendUnixOpticalDrives(drives);
+                drives = AppendUnixFloppyDrives(drives);
```
mnadareski: The reason that floppy devices should be locked behind the fixed drives flag is because not every program supports it by default, very notably Redumper does not. Realistically, this should impact a very small number of users, but having it maintain the same behavior as Windows would be appreciated.

MPF.UI/Windows/MainWindow.xaml
```diff
@@ -209,7 +209,7 @@
                         </ComboBox.ItemContainerStyle>
                     </ComboBox>
                     <ComboBox x:Name="MediaTypeComboBox" Grid.Row="0" Grid.Column="1" Height="22" Width="140" HorizontalAlignment="Right"
-                              ItemsSource="{Binding MediaTypes}" SelectedItem="{Binding Path=CurrentMediaType, Converter={StaticResource ElementConverter}, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}"
+                              ItemsSource="{Binding MediaTypes}" SelectedItem="{Binding Path=CurrentPhysicalMediaType, Converter={StaticResource ElementConverter}, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}"
```
mnadareski, line 212: Good catch, thank you!

MPF.Frontend/Drive.cs
gmipf: Done in f6e15eac — floppy enumeration now runs inside the `!ignoreFixedDrives` block (before the fixed enumerator so a USB floppy is still tagged Floppy and skipped there), so it matches the Windows behavior where floppies only appear once fixed drives are shown. Comment updated to note the Redumper rationale. 558/558 tests still pass.

mnadareski (approved):

PR #1007, opened 2026-07-07
Pre-compress subcode files with Zstd

Deterous:
Similar to state and skeleton, pre-compressing subcode file can have significant savings for the log zip size for some CDs

### DIFF
--- a/CHANGELIST.md
+++ b/CHANGELIST.md
@@ -1,3 +1,6 @@
+### WIP (xxxx-xx-xx)
+- Pre-compress subcode files with Zstd
+
 ### 3.8.3 (2026-07-06)
 
 - Add live tool output console for the Linux GUI (gmipf)
--- a/CHANGELIST.md
+++ b/CHANGELIST.md
@@ -1,4 +1,5 @@
 ### WIP (xxxx-xx-xx)
+
 - Pre-compress subcode files with Zstd
 
 ### 3.8.3 (2026-07-06)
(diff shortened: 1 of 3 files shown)


MPF.Processors/Redumper.cs
```diff
@@ -571,7 +573,7 @@ internal override List<OutputFile> GetOutputFiles(PhysicalMediaType? mediaType,
                         new($"{outputFilename}.state.zst", OutputFileFlags.Binary
                             | OutputFileFlags.Zippable,
                             "state_zst"),
-                        new($"{outputFilename}.subcode", OutputFileFlags.Required
+                        new([$"{outputFilename}.subcode", $"{outputFilename}.subcode.zst"], OutputFileFlags.Required
```
mnadareski: These should be separate lines like `.state` and `.state.zst`

CHANGELIST.md
```diff
@@ -1,3 +1,6 @@
+### WIP (xxxx-xx-xx)
+- Pre-compress subcode files with Zstd
```
mnadareski, line 3: Nitpick: Newline before the first item in the list

MPF.Processors/Redumper.cs
Deterous: How do you manage Required attribute when you need one of either of them

mnadareski: Given that the subcode file isn't used by any of the processing code, it's really not "required". In previous cases, the required flag was removed from both. I suppose if this handles all 3 potential zipping cases (2 separate or both existing), then this is fine and the others might follow suite.

Deterous: I don't know how the two-possible-names situation like this actually works

mnadareski: If you feel strongly about this keeping the required flag, then let's find out!

Deterous: I guess it's not *required* so can just remove that

mnadareski (approved):

mnadareski (approved):

PR #1010, opened 2026-07-12
Add macOS support to the GUI

gmipf:
MPF finds no drives at all on macOS: the GUI shows "Found no drives" and no dump can be started.

`Environment.OSVersion.Platform` reports `PlatformID.Unix` on macOS, so `GetDriveList` takes the Linux path, whose enumerators read `/sys/block`, `/sys/class/scsi_generic` and `/dev/sr*`. None of those exist on macOS, so each returns nothing without failing and the drive list comes back empty.

Four commits, each one thing.

### Add macOS drive enumeration

A macOS enumerator built on `diskutil info`, mirroring the structure of `Drive.Linux.cs`.

Optical drives are identified by the **presence** of diskutil's optical-only fields rather than by any value they carry: an emulated drive leaves "Optical Media Type" blank and reports its protocol as SATA, while a real USB drive fills the field in and reports USB, so neither value is a reliable marker on its own.

Optical media is left out of the `DriveInfo` query here. macOS mounts every readable disc on insertion, and `DriveInfo` reports the mounted volume (for example `/Volumes/LABEL`) as a CD-ROM drive. A mount point is not a dumpable device, and diskutil already surfaces the drive holding the disc, so the volume would only add a second, unusable entry for the same disc.

### Name macOS drives by BSD name

The drive name reaches the dumping program verbatim. On macOS a program takes a drive as its BSD name rather than as a device node path: redumper resolves the name against the BSD name IOKit publishes for the drive and fails with

    failed to find matching SCSI authoring device

when it is given `/dev/disk4`. Without this, every dump from an enumerated drive fails.

This mirrors the existing platforms: the name is a drive letter on Windows and a device node on Linux, in both cases the form the program expects.

### Remove the Redumper drive workaround on macOS

`PrepareMacRedumperParameters` stripped `--drive` from the Redumper parameters whenever the selected drive name began with `/Volumes/`. It existed because the mounted volume was the only optical entry macOS used to surface, and no dumping program takes a mount point as a drive.

Dropping the argument made Redumper fall back to picking the first drive that holds a disc, which is only ever right when a single drive is loaded. With two loaded drives it silently dumped one the user had not selected.

With the device itself enumerated and passed as its BSD name, the workaround has nothing left to match and the guess is no longer needed.

### Show the tool output console on macOS

A dumping tool gets no console window of its own on macOS, the same as on Linux, so the GUI ran the tool with no visible output at all.

The console window added for the Linux GUI (#998) already fits: the window, its terminal buffer, and the output pump are platform-neutral. Only the gate that creates the console was limited to Linux, so macOS is added to it. This is the whole change:

```diff
-            IToolOutputConsole? toolConsole = OperatingSystem.IsLinux()
+            IToolOutputConsole? toolConsole = OperatingSystem.IsLinux() || OperatingSystem.IsMacOS()
```

### Verification

macOS 26 (x86_64) with a USB PLEXTOR PX-760A. The drive is enumerated, selected in the GUI, and dumped end to end with redumper: a mixed-mode CD, 0 SCSI errors, 0 C2 errors, and a `.scram` byte-for-byte identical in size to the same disc dumped on Linux. The tool output console shows the live output during the dump.

316 lines of tests are added to `MPF.Frontend.Test/DriveTests.cs`, covering the diskutil parsing, the optical-drive detection, and the BSD naming. The suite is green (582 tests).

---


MPF.Frontend/MacOSBlockDevice.cs
```diff
@@ -0,0 +1,38 @@
+namespace MPF.Frontend
+{
+    /// <summary>
+    /// A macOS whole-disk device discovered via diskutil.
+    /// </summary>
+    internal sealed class MacOSBlockDevice
```
mnadareski: In changes that were queued up while this was being made, I removed the UnixBlockDevice type that this was modelled on. I would strongly prefer using `Drive` for all appropriate places instead of creating another new type that essentially does the same thing.

MPF.Frontend.Test/DriveTests.cs
```diff
@@ -396,5 +396,321 @@ private static void WriteSysBlockDevice(string root, string name, string removab
...
+        }
+
+        [Fact]
+        public void EnumerateMacOSWholeDiskNodes_OnlyMatchesDiskFollowedByDigits()
+        {
+            string root = Path.Combine(Path.GetTempPath(), "mpf-test-dev-" + Guid.NewGuid().ToString("N"));
+            Directory.CreateDirectory(root);
+            try
+            {
+                File.WriteAllText(Path.Combine(root, "disk0"), string.Empty);
```
mnadareski: When I moved some of the previous methods to SabreTools.IO, I realized that each of these tests was constantly creating and deleting the exact same files every time. An easier and cheaper way of doing this is having the test files setup and just copied so the test runner can read them as they are without having to worry where they might end up. Look for instances of `TestData` in tests and project files to get an idea of what this would look like.

MPF.Frontend/Drive.cs
```diff
@@ -286,12 +286,21 @@ public void RefreshDrive()
...
-            var desiredDriveTypes = new List<DriveType>() { DriveType.CDRom };
+            // macOS mounts every readable disc on insertion, so DriveInfo reports the mounted
+            // volume (e.g. "/Volumes/LABEL") as a CD-ROM drive. That is a mount point rather
+            // than a dumpable device, and diskutil already surfaces the drive holding the disc
+            // as a device node, so optical media is left out of the DriveInfo query there too.
+            bool isMacOS = IsMacOS();
+
+            var desiredDriveTypes = new List<DriveType>();
+            if (!isMacOS)
+                desiredDriveTypes.Add(DriveType.CDRom);
```
mnadareski: I feel like keeping the original logic here is pretty harmless even if it's not used in the MacOS path.

mnadareski (commented):
There are definitely going to be merge conflicts given the amount of changes I did.

MPF.Frontend/MacOSBlockDevice.cs
gmipf: Done — `MacOSBlockDevice` is gone. `ParseMacOSDiskutilInfo` now returns a `Drive` and `EnumerateMacOSDevices` a `List<Drive>`, built with an object initializer the same way `EnumerateUnixFixedDevices` does. The BSD name is the `DevicePath`, which is what `DumpEnvironment` hands to the dumping program verbatim.

MPF.Frontend.Test/DriveTests.cs
gmipf: Done — the seven fake device nodes now live in `MPF.Frontend.Test/TestData/MacOSDev` and are copied to the output directory, so the test just reads them where they are. I checked that it still fails without them (`Assert.Equal() Failure: Collections differ`) rather than passing vacuously.

The sysfs tests above still build their trees at runtime. Happy to convert those in this PR too if you want them done together.

MPF.Frontend/Drive.cs
gmipf: Reverted — `GetDriveList` is back to `new List<DriveType>() { DriveType.CDRom }`.

It does have an effect on macOS though, so rather than dropping the handling I moved it into the macOS enumerator. macOS automounts every readable disc, and .NET reports that mount as a `DriveType.CDRom` drive, so the mount point ends up in the list next to the device that holds the disc. Measured just now on macOS 26.5, with the disc mounted (`/dev/disk1s0 on /Volumes/MPFTEST`):

Without the filter:

```
DevicePath=/Volumes/MPFTEST  Type=Optical  Display=MPFTEST
DevicePath=disk1             Type=Optical  Display=disk1
DevicePath=disk4             Type=Optical  Display=disk4
```

With it:

```
DevicePath=disk1             Type=Optical  Display=disk1
DevicePath=disk4             Type=Optical  Display=disk4
```

The `/Volumes/...` entry is not dumpable: `DumpEnvironment` hands `DevicePath` to the program verbatim, and Redumper wants the BSD name. So it is a second, failing entry for a disc that is already listed under `disk1`. `AppendMacOSDrives` now drops mount points, which is also what `PrepareMacRedumperParameters` was working around — that is why this branch removes it.

For what it is worth, Linux produces the same double entry on an auto-mounting desktop (`/run/media/<user>/MPFTEST` next to `/dev/sr0`). That is pre-existing and left alone here.

mnadareski (approved):

PR #1028, opened 2026-09-10
Check UI drive dropdown

### LINKED ISSUE #965: [Request] Add drive dropdown to Check UI
Add _optional_ drive dropdown to Check UI. This needs to differ from the main UI as providing a disc path is optional for Check in general. This would then mirror the behavior of Check CLI.

Deterous:
Fixes #965 
Drive dropdown defaults to None and if a drive is selected it will use that drive during Check and scan it.
Pic of UI below:
image
And while scanning:
image

### DIFF
--- a/MPF.Avalonia/Windows/CheckDumpWindow.axaml
+++ b/MPF.Avalonia/Windows/CheckDumpWindow.axaml
@@ -24,7 +24,7 @@
                     <TextBlock Text="{DynamicResource SettingsGroupBoxString}"
                                FontWeight="Bold" />
                     <Grid ColumnDefinitions="160,*,Auto"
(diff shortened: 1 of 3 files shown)


mnadareski:
General question: would it be possible to also wire up the Avalonia side? It should be essentially the same code.

Deterous:
> General question: would it be possible to also wire up the Avalonia side? It should be essentially the same code.

Possible? Yes. Do I know how? No.

mnadareski:
If you can give it a shot, I would appreciate it. If you can't get something compiling quickly, then it can be added at a later date.

Deterous:
Not tested, but if it builds, it builds

mnadareski (approved):

